PR #99, opened 2021-01-26
Add environment variables in dockerfile for tests to pass

angelhof:


nvasilakis (approved):
I believe line 15 replaces line 12, right? (also: @GNtousakis ?)

GNtousakis:
Yeap that seems right! 
ENV variable should work the same as run export.

PR #102, opened 2021-01-27
Adjust scripts/remote to support many workers

zyrolasting:
This revision adjusts the configuration model of `scripts/remote/controller` such that you can start multiple controllers that each manage their own host. Some research yesterday showed that a single process approach would take longer to develop, and end up with relatively poor fault tolerance. A multi-process approach only required more settings, and some code to move to different files.

There is one breaking change: This removes endpoints like `/ci` and `/perf-ci` because that assumes a single-process model. Each controller only needs `/run` in a multi-process model. We can still have the named endpoints on the NGINX-side by redirecting to the correct `/run`.

This is an example of what starting multiple controllers would look like.

```
$ cd pash/scripts/remote/controller
# This one coordinates correctness tests
$ PASH_REMOTE_RCFILE=configs/correctness-tests.json node main.js &
# This one would coordinate performance tests
$ PASH_REMOTE_RCFILE=configs/performance-tests.json node main.js &
...
$ curl localhost::2047/run # start correctness tests
$ curl localhost::2048/run # start performance tests
$ curl localhost::2047/now # correctness status
$ curl localhost::2048/now # performance status
```


scripts/remote/README.md
```diff
@@ -102,20 +112,43 @@ powerful) machine you want to run the script from that point on. Copy
...
+[cm]: #command-modules
+
+`commands/default.js` simply runs the worker script without waiting
+for it to complete.
+
+`commands/correctness.js` assumes the worker script is
+`workers/run-ci.sh`, or a script based on it. If the controller
+detects relevant output, it downloads a `reports` directory from the
+remote host. *This is for backwards compatibility to make sure reports
+are visible to the NGINX server using `scripts/pash-nginx.conf`*.
```
nvasilakis, line 125: Idea: would it make sense to store this in the config? e.g., we could have two entries for correctness and performance reports, rather than hardcoding it in the command.

scripts/remote/README.md
```diff
@@ -62,7 +78,8 @@ e.g. `host` -> `PASH_REMOTE_HOST`, or `my-conf1g-key` -> `PASH_REMOTE_MY_CONF1G_
 
 ### API
 
-* `/run` runs `~/work.sh`.
+* `/run` runs the worker's `~/work.sh`, with added invariants
+  depending on the server's [command module][cm].
```
nvasilakis, line 82: Is `[cm]` intended to be `(cm)` (if for Markdown link?) . 

(You can also link directly to the subsection title using `(#command-modules)` :)

scripts/remote/controller/.gitignore
```diff
@@ -1 +1,2 @@
 reports
+rc.json
```
nvasilakis, line 2: Great!

nvasilakis (approved):
Just to confirm, is the idea here to start a new controller per command module? Or could we have two command modules (reading two different JSON files in `configs`) running in the same controller process? (I'm almost certain you're doing the second, just wanted to confirm).

nvasilakis:
Sorry, I'm just seeing you're going with the multi-process model — I had missed that!

scripts/remote/README.md
nvasilakis: Update: We're already changing the nginx config to point to different processes, so feel free to break compatibility here too:-)

scripts/remote/README.md
zyrolasting: This is valid syntax for [full-reference style links](https://github.github.com/gfm/#full-reference-link). See line 116 for the named reference to `#command-modules`.

scripts/remote/README.md
zyrolasting: I'm not sure what you mean here. Store what in the config?

zyrolasting:
>  could we have two command modules (reading two different JSON files in configs) running in the same controller process?

We can, with some refactoring. But note that if we do this, and the process goes down, no one is running any tests until the problem is fixed. The way it is now, one process going down only affects the tests covered by that process. I suggest we leverage that.

scripts/remote/README.md
nvasilakis: Thanks for this!

scripts/remote/README.md
nvasilakis: I meant the same "this" as the one in the comment on line 124, which I assumed to mean "the directory where the reports go". And I meant (i) this could be part of the config, (ii) we can easily change where `pash-nginx.conf` looks for them.

nvasilakis (approved):

PR #103, opened 2021-01-28
Add a superoptimizer hook

nvasilakis:
This creates a quick-and-easy way to plug in the superoptimization work. The flag will eventually find its way into the Python side, but for now this calls a `compiler/superoptimize.sh` script that could serve as a starting point.


nvasilakis:
great catch!

angelhof (approved):

PR #104, opened 2021-01-28
Prefer UTF-8 encoding with read and generated UNIX50 files

zyrolasting:
`generate_unix50_scripts.py` fails with the following stack trace for `evaluation/unix50/8.txt`, when the system prefers the `ascii` codec.

```
Traceback (most recent call last):
  File "generate_unix50_scripts.py", line 44, in <module>
    input_file_data = file.read()
  File "/usr/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 28011: ordinal not in range(128)
```

`cat unix50/8.txt | dd ibs=1 skip=28010 count=3` shows that the `ascii` codec tried to decode a Unicode cross character, which appears in a sentence about chessboard dimensions.

The root cause is that `open` will use a platform-specific default encoding, and `ascii` may be used in the context of a container. This seems incorrect given the input data, hence this patch.

I confirmed that with this patch, `execute_unix_benchmarks.sh` runs to completion when the platform prefers the `ascii` codec.

### DIFF
--- a/evaluation/eurosys/generate_unix50_scripts.py
+++ b/evaluation/eurosys/generate_unix50_scripts.py
@@ -39,12 +39,12 @@ input_file_names = input_file_name_assignments.values()
 for input_file in input_file_names:
     input_file_path = os.path.join(unix50_dir, input_file)
     generated_input_file_path = os.path.join(generated_inputs_dir, input_file)
-    with open(input_file_path) as file:
+    with open(input_file_path, encoding='utf-8') as file:
         input_file_data = file.read()
 
     # print("Input:", input_file_path, "size:", len(input_file_data))
     num_iterations = maximum_input_size // len(input_file_data)
-    with open(generated_input_file_path, "w") as file:
+    with open(generated_input_file_path, "w", encoding='utf-8') as file:
         for _ in range(0, num_iterations, 100):
             file.write(input_file_data * 100)
 


nvasilakis (approved):
Ah great catch — looks good!

angelhof:
Thanks, is this merge urgent? If not I would propose delegating merging it since it might affect EuroSys artifact results.

If it is urgent I can take a look and do a best effort pass to ensure that results are mostly similar.

zyrolasting:
@angelhof I can't run the tests in my container unless I figure out how to make the container communicate to Python that it prefers UTF-8.

If you check the container on your dev machine, what codec does `open` prefer for you? If it uses `utf-8`, then all this code does is make what was already happening explicit.

nvasilakis:
I'm wondering if we can setup the proper environment variables in the container. What comes to mind is:

```
LANG="en_US.UTF-8"
LANGUAGE="en_US.UTF-8"
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_PAPER="en_US.UTF-8"
LC_NAME="en_US.UTF-8"
LC_ADDRESS="en_US.UTF-8"
LC_TELEPHONE="en_US.UTF-8"
LC_MEASUREMENT="en_US.UTF-8"
LC_IDENTIFICATION="en_US.UTF-8"
LC_ALL="en_US.UTF-8"
```

@zyrolasting are you using the build file, are you pulling [img.pash.ndr.md](img.pash.ndr.md), or are you doing something else?

zyrolasting:
@nvasilakis I've always used `img.pash.ndr.md`.

zyrolasting:
@nvasilakis Also, what you are suggesting does the exact same thing in the Python runtime...

angelhof:
Great, I checked on the development machine using `locale.getpreferredencoding()` and I got `UTF-8` so your change is probably fine!

angelhof (approved):

mgree:
`export PYTHONIOENCODING=utf8` if you need it, but seems like things are good. :)

PR #111, opened 2021-01-31
fix more hardcoded paths to /bin/bash -> /usr/bin/env bash

kidd:
Here are a couple of paths that needed to be fixed to run the basic pash examples.

Grepping around I noticed lots of others.

Maybe it'd be good to migrate them over to `!#/usr/bin/env bash` too.

### DIFF
--- a/compiler/pash.py
+++ b/compiler/pash.py
@@ -89,7 +89,7 @@ def preprocess(ast_objects, config):
     return preprocessed_asts
 
 def execute_script(compiled_script_filename):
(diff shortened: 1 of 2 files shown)


nvasilakis:
This looks good, thanks a lot!

angelhof (approved):
Thanks a lot!

Tested and looks good. I guess that at some point we could do a proper effort to replace all `/bin/` with `/usr/bin/env` but I would defer it for now until the need arises for someone :)

kidd:
sure. I'm replacing them only on a need basis (more back and forth with reviews, but higher chances of merge)

Issue #108, opened 2021-01-29
Make performance test results consistent


zyrolasting:
It's hard to tell what's going on in the performance test results.

* Some of the file names appear to have a relationship with `--width`, e.g. `wf_8_distr.time`, but then I see things like `wf_50_distr.time` or `p3b4_10_distr.time`.
* Underscores seem to play multiple roles in file names. If you split `wf_64_distr.time` by underscores, the test name is clearly the first element of the result. But in `web-index_p3_2gram_2_distr.time`, the test name is the first... 3 elements?
* Some files appear abandoned in version control. e.g. `_distr.time`, maybe some subdirectories?
* Some `*.time` files have more than just the output of `time`. They frequently have STDERR from `kill`, or messages like `Files ../evaluation//intermediary//wf_seq_output and /tmp/distr_output/0 are identical`.

I don't have a strong preference for the result format, but it would help if both the file names and contents were consistent.

angelhof:
Thanks for these comments!

> Some of the file names appear to have a relationship with --width, e.g. wf_8_distr.time, but then I see things like wf_50_distr.time or p3b4_10_distr.time.

Some of these are very old files and results. I will delete them :)

> Underscores seem to play multiple roles in file names. If you split wf_64_distr.time by underscores, the test name is clearly the first element of the result. But in web-index_p3_2gram_2_distr.time, the test name is the first... 3 elements?

That's a good point, I guess from now on we can have a convention (for the old files I am parsing from the right).

> Some *.time files have more than just the output of time. They frequently have STDERR from kill, or messages like Files ../evaluation//intermediary//wf_seq_output and /tmp/distr_output/0 are identical.

This is a very good point. Actually all `.time` files have the pash logging information so that we can extract information from it. For example, PaSh outputs more fine grained time information (separated for compilation, optimization, execution, etc). Until know we have been using that information to plot performance results. Maybe from now on we can fix this to be slightly different, however I think that we might need pash logging information again. Do you have any ideas of how to solve this? Maybe we can just change the output file names to `.log` instead of `.time`?

zyrolasting:
New bullet:

* `baseline_sort` has `_seq` and `_small_seq` variants whereas other datasets have only `_seq`.

PR #113, opened 2021-02-01
Analyze/report performance test results 

zyrolasting:
This code reads performance test results, and can optionally present a tabular summary. This functionality is not yet integrated into a CI system, so I need a review over how I'm extracting data (Empirical means are necessary due to  #108)

e.g.

```
╔════════╤════════════╤════════════╗
║        │ topn       │ wf         ║
╟────────┼────────────┼────────────╢
║ abd611 │ 473277.432 │ 179270.421 ║
╚════════╧════════════╧════════════╝
```

(The borders come from NPM's [table][] package, and are easily removed)

[table]: https://www.npmjs.com/package/table
 
To reproduce the above table, do this:

```
const { analyze, keepWidth, keepTimed, render } = require('./perf-analysis.js');

// change me
const resultsPath = '/path/to/pash/evaluation/results';

const data = analyze(resultsPath);

// Keep only the timed results given --width 64
const filtered = keepTimed(keepWidth(data, 64)).filter(({variant}) => variant === 'distr');

// render() takes the data in an array like [['commit-hash', { ... }], ['commit-hash', { ... }]]
console.log(render([['abd611', filtered]], ['wf', 'topn']))
```

EDIT: Updated filter in example to show only `_distr.time` timings.


scripts/remote/controller/perf-analysis.js
```diff
@@ -0,0 +1,93 @@
...
+              variant: extractVariant(absolutePath),
+          };
+      });
+
+// Pash --width is in the file name
+const extractWidth = (p) =>
+      parseInt((path.basename(p, path.extname(p)).match(/_(\d+)_/) || [])[1]);
+
+const extractVariant = (p) =>
+      (p.match(/_distr_?([^\.]+)/) || [])[1];
```
zyrolasting, line 41: I can't issue a change request on my own PR, so I'll just flag this here.

This pattern looks for test variants that strictly follow `_distr_?`, which means `_seq` is not seen as a test variant. It might be better to parse from right to left as @angelhof mentioned in #108.

angelhof (approved):
This looks good to me! One note worth considering is that `_distr_auto_split.time` are the _best_ PaSh results. When they don't exist the reason is usually that `_distr.time` corresponds with `_distr_auto_split.time`. However I don't think that this should be a concern for now.

> so I need a review over how I'm extracting data

What exactly is the issue now with data extraction?

zyrolasting:
> What exactly is the issue now with data extraction?

It's just that I'm yanking out values from text logs using regular expressions. That, and you were mentioning some of the results are old. Altogether the process felt error prone, so I wanted to hear if the values you can produce using this code are consistent with your expectations.

angelhof:
OK got it! 

There are a couple points that need to be addressed to get the table results that we actually want.
- There could be multiple "Execution time:" lines in a log file, so we would want to sum them to get correct results. Alternatively, we could also parse the total time (produced from the time command) which includes all of the PaSh overhead (compilation, optimization, parsing, etc). Maybe we want both :)
- We would usually go for speedup. To show that we would need to get the `*seq.time` time (which is produced by `time` and not PaSh so it has to be parsed differently) and divide it by the PaSh execution time. Then we can report the sequential time next to the speedup (e.g. seq-time: 100s (x10 pash-speedup))
- For the most up to date results (that we would want to visualize in the table), we would need to look at the perf-benchmarks that are run now. I think that you are running `evaluation/eurosys/execute_eurosys_one_liners.sh -m` and `evaluation/eurosys/execute_unix_benchmarks.sh -s`. The results for the first are stored in `evaluation/results/eurosys_small/` directory and the files to look for PaSh are the `distr_auto_split.time` files. The results for the second are stored in `evaluation/results/unix50_4_1073741824` again in the `distr_auto_split.time` files.
- For now, I would fix a width (e.g. 16) for each benchmark set, and the variant, and just visualize them in a commit by commit table as follows (the numbers are made up):

```
Results for width 16, full PaSh (distr_auto_split), eurosys-one-liners:
╔════════╤═════════════╤═════════════╗
║        │ topn        │ wf          ║
╟────────┼─────────────┼─────────────╢
║ abd611 │ 473s, x10.3 │ 270s, x5.4  ║
╚════════╧═════════════╧═════════════╝
```

zyrolasting:
@angelhof Thank you for the excellent and detailed feedback! I'm already on the way with last two bullets you mentioned after convos with @nvasilakis, so we're good there. Just a couple of follow up questions:

> Summing execution time

Using vanilla JS I'd have to parse these as floats and add them together. Is floating point imprecision tolerable here? Would I need to report an error margin alongside the tables?
 
> Maybe we want [multiple times]

Is that something that has to be implemented in this PR?

angelhof:
Great!

> Using vanilla JS I'd have to parse these as floats and add them together. Is floating point imprecision tolerable here? Would I need to report an error margin alongside the tables?

 I think floating point is fine (that is what I have been using until now too).

> Is that something that has to be implemented in this PR?

I don't think so, I just mentioned it for more context and as a possible extension later :)

zyrolasting:
@angelhof I updated the code to handle the requirements you mentioned. I needed this branch for future work anyway, so please forgive my asking for a second review pass. 

* Added README
* Includes a utility module (`history.js`) I'll need to integrate this into CI later
* Now reporting the sum of all PaSh execution times in a file.
* I figured out a way to report the data contextually. The below screenshots show that timing information is presented based on available data.
    - Sequential tests: `real`, `user`, `sys` timings.
    - Dist. tests + variants, Runtime in seconds, with speedup factor if adjacent `_seq.time` is available 

(*NOTE:* I intentionally left out the commit column here. It's trivial to add back once I have that data ready.)

image
image

angelhof (approved):

angelhof:
> @angelhof I updated the code to handle the requirements you mentioned. I needed this branch for future work anyway, so please forgive my asking for a second review pass.

No worries! Thanks a lot for the changes :)

Everything looks good for now. I think we should merge after Nikos reviews.

At some later point we can discuss about presentation details and inconsistencies regarding the timing data that is produced by PaSh and hopefully clean up and rebuild the experiment running infrastructure to have consistent folder/file names.

nvasilakis:
@zyrolasting Thanks for all your work here!

What is the purpose of `history.js`? Is it manipulate results from all commits as a little JSON DB?

More generally, a simple solution here would be to copy the behavior from the "correctness" reports, but just change the data you're using and their presentation: the same way we have multiple columns there with different _correctness_ numbers, we would simply have different columns here with different _performance_ numbers. Summaries would be flat append-only logs (and optionally per-commit files would have more details about the performance result of a single commit — but this can come later).

zyrolasting:
> What is the purpose of history.js? Is it manipulate results from all commits as a little JSON DB?

In the context of performance tests, yes. I mentioned this in my other comment.

Issue #97, opened 2021-01-26
Move all logs to user-specific directory, rather than /tmp


nvasilakis:
Currently, Pash writes all logs and intermediate files to `/tmp`. This causes a problem when many different users attempt to use PaSh on the same computer — the common case in shared large-scale multiprocessors. One idea would be to quickly change all `/tmp` pointers to some `$HOME`-local directory. A better option would be to pair this with a `pash --init` flag that is used to set up Pash.

(Additionally, jobs could be given IDs so that they do not mess up with each other when multiple invocations of PaSh operate in concurrently.)

angelhof:
Addressed since logs are written in invocation specific pash direcotry.

PR #115, opened 2021-02-02
nest tempfiles into a unique tempdir

kidd:
- This PR tries to unify all tmpfiles to live inside a unique directory.
- Some trailing whitespace cleaning
- runtime scripts are still outside the tempdir but I'm not sure it's that simple to move it without messing up :)

### DIFF
--- a/compiler/ast_to_ir.py
+++ b/compiler/ast_to_ir.py
@@ -252,7 +252,7 @@ def compile_node_command(ast_node, fileIdGen, config):
             compiled_ast = make_kv(construct_str,
                                    [ast_node.line_number, compiled_assignments,
                                     compiled_arguments, compiled_redirections])
-            
+
         return compiled_ast
 
 def compile_node_and_or_semi(ast_node, fileIdGen, config):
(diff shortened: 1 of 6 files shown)


angelhof:
Thanks a lot, this is a great change, going towards the right direction :)

I think we should use this PR to (1) also address #97 and (2) make sure that runtime output is also in this dir. This would allow for multi-user execution.

### Regarding (1):

The easiest way to go would be  to modify `pa.sh --init` (https://github.com/andromeda/pash/blob/main/pa.sh#L7) to create a directory `~/.pash/`. Then `pa.sh` could complain if that directory doesn't exist and it could ask for the user to run `pa.sh --init`. 

Then all of the temp files could be created in `~/.pash/temp/` (which should be a very easy change since you have gathered all logic for creating temp files in one place.

### Regarding (2):

At the moment the runtime creates temp files using `"$(mktemp --tmpdir -u pash_XXXXXXXXXX)"` (only used in `compiler/pash_runtime.sh` and `compiler/pash_runtime_quick_abort.sh` if I remember correctly). However, since we are going down that path, we could create a `pash_temp_file_name.sh` that calls your python code that gets a tempfile name and then use that in all places in the runtime to ensure that this logic is only in one place.

If you have time and are interested in doing these changes that would be great, otherwise I can look into them when I have some more time!

kidd:
Hi! thanks for the feedback.

1. not sure the user wants to specifically run a step to initialize, if it's possible to run transparently.  in this case, the path is a conf. I'd say a sane approach is to default to that tempdir, that will effectively create a new directory on every run (which is cool for debugging purposes, if you add a condition to not rm it if `-d 1`). and if the user wants a particular one, then let him put it in a config.yaml (that I didn't implement). 

2. `man mktemp` says that if there's an env var TMPDIR, it'll use it. idk if this is too intrusive for what you want (not sure if the compiled scripts will want to have that overriden, but anyway, this only works when you don't specify any explicit dir in mktemp, so you usually don't care in those cases).

Feel free to merge and modify it to your needs.

angelhof:
Thanks! I will take a look and apply my changes in the next couple days.

angelhof:
I am closing this since it is subsumed by #116

PR #116, opened 2021-02-06
Nest all pash temp files under the same directory.

angelhof:
Subsumes #115 

PaSh now nests all files under the same directory.

It also creates a different directory per execution.

Finally, it deletes this temporary directory if not under debugging mode.

### DIFF
--- a/compiler/pash.py
+++ b/compiler/pash.py
@@ -92,7 +92,7 @@ def preprocess(ast_objects, config):
 
 def execute_script(compiled_script_filename):
     new_env = os.environ.copy()
-    new_env["TMPDIR"] = config.PASH_TMP_PREFIX
+    new_env["PASH_TMP_PREFIX"] = config.PASH_TMP_PREFIX
     exec_obj = subprocess.run(["/usr/bin/env", "bash" ,compiled_script_filename], env=new_env)
     ## Return the exit code of the executed script
     exit(exec_obj.returncode)
--- a/compiler/pash.py
+++ b/compiler/pash.py
@@ -5,7 +5,6 @@ from datetime import datetime
 
 from annotations import *
 from ast_to_ir import *
-from pash_runtime import *
 from ir import *
 from json_ast import *
 from parse import parse_shell, from_ir_to_shell_file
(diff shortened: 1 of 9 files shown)


compiler/pash.py
```diff
@@ -67,29 +70,34 @@ def parse_args():
...
 
-def execute_script(compiled_script_filename):
-    exec_obj = subprocess.run(["/usr/bin/env", "bash" ,compiled_script_filename])
+def execute_script(compiled_script_filename, debug_level):
+    new_env = os.environ.copy()
+    new_env["PASH_TMP_PREFIX"] = config.PASH_TMP_PREFIX
+    exec_obj = subprocess.run(["/usr/bin/env", "bash" ,compiled_script_filename], env=new_env)
+    ## Delete the temp directory when not debugging
+    if(debug_level == 0):
+        shutil.rmtree(config.PASH_TMP_PREFIX)
```
kidd, line 100: appart from checking debug_level, it'd be good to make sure that PASH_TMP_PREFIX was not provided by the user, but created ourselves. It's not 

`PASH_TMP_DIR=~/ pash.sh` could have interesting consequences :)

angelhof: Thanks, very good point :) I modified it so that `pash.py` (the entrance for pash) always sets this itself (and only the runtime that doesn't do any deleting inherits it).

kidd: I don't fully get the cycle of pash compile vs runtime.... but if it works fine... that's ok. 

some tests wouldn't hurt either :)

angelhof: At the moment the standard tests (in `compiler/test_evaluation_scripts.sh`) are OK witnesses of correctness. I tried to think about some test to check that directories are deleted correctly, but it would be very elaborate for what it checks I think.

So I guess that we could merge if you don't have any idea of a straightforward to do test to check that. Do you have any other property to check in mind?

nvasilakis (approved):
Looks good to me!

nvasilakis:
Thanks @kidd, @angelhof !

compiler/pash.py
angelhof: I am merging the PR and let's see if we want additional tests later!

Issue #118, opened 2021-02-08
CI code should be in new repo


zyrolasting:
Since CI code is hosted in the same repository, it will run into situations like this:

```
git checkout $commit_to_test
run_tests
git checkout $commit_with_relevant_script
process_results
```

This is especially troublesome because it creates code [that would break on merge](https://github.com/andromeda/pash/blob/a0b1feb25d40ccb7c59fa5f5c74b11d65ee4ee11/scripts/remote/workers/run-perf.sh#L22).

I suggest hosting CI code separately to speed up development a bit. It takes more time to babysit something like this.

angelhof:
That is a good point. Also I guess that CI related stuff might be better off in a private repo. @nvasilakis what do you think?

nvasilakis:
This has been discussed offline, and we decided to move CI-related things into a new repo.

PR #119, opened 2021-02-08
Summarize performance data

zyrolasting:
Due to #118 and my current permissions, I can only say this works off my branch when I ping memstar directly (other factors made this tricky to implement). I will need to verify functionality again once this merges to `main`, and the NGINX config gets reloaded.

Once that happens:

* `perf.pash.ndr.md` serves summary data.
* `perf-ctrl.pash.ndr.md/run` runs the defined performance tests. `/stdout`, etc. work as before.


scripts/Dockerfile
```diff
@@ -16,4 +16,18 @@ ENV LANGUAGE en_US.UTF-8
...
+ENV NODE_VERSION 8.9.4
+RUN curl https://raw.githubusercontent.com/creationix/nvm/v0.37.2/install.sh | bash \
+    && . $NVM_DIR/nvm.sh \
+    && nvm install $NODE_VERSION \
+    && nvm alias default $NODE_VERSION \
+    && nvm use default
+
+ENV NODE_PATH $NVM_DIR/versions/node/v$NODE_VERSION/lib/node_modules
+ENV PATH      $NVM_DIR/versions/node/v$NODE_VERSION/bin:$PATH
+RUN cd /pash/scripts/remote/controller && npm install && cd -
```
angelhof, line 31: Would it make sense to have those in a separate Dockerfile (only for the CI and controller)?

angelhof (approved):
Except for the Dockerfile comment this looks good to me.

scripts/Dockerfile
zyrolasting: If we are going for a separate repo for CI, then yes.

zyrolasting: My only hesitation is that the second Dockerfile would copy most of the first anyway.

angelhof: Hmm, I see. I guess I was thinking that the controller for the perf experiments is not included in the default environment for the project (a user can still run the perf experiments on their own machine without having node).

If however `node` does not increase docker build time I guess that we could leave it in as is.

angelhof: This exists too: https://docs.docker.com/develop/develop-images/multistage-build/

scripts/pash-nginx.conf
```diff
@@ -16,6 +20,18 @@ server {
...
+	server_name perf.pash.ndr.md;
+	access_log /var/log/nginx/pash.access.log;
+  error_log  /var/log/nginx/pash.error.log;
+  default_type text/plain;
+	location / {
+		root   /var/www/pash-web/results/;
+		index  index;
+	}
+}
+
```
nvasilakis, line 34: awesome, thanks!

scripts/pash-nginx.conf
```diff
@@ -74,6 +90,25 @@ server {
   }
 }
 
+server {
+  listen 80;
+  server_name perf-ctrl.pash.ndr.md;
```
nvasilakis, line 95: Not for now, but for later:

The `ctrl.pash.ndr.md` could remain the same—pointing to the same controller script—but just have a few additional endpoints for performance (and locks). Generally having a single controller is ideal, because we can handle all queuing / concurrency / barrier logic at a single place.

nvasilakis (commented):
This looks good for now, thanks a lot!

Later, based on our discussion, summarization scripts, including `parse.js` and `report.js`, can be implemented as trivial shell scripts and remove any node dependency (except for the server/controller).

PR #128, opened 2021-02-16
poets scripts for pash

dkarnikis:


evaluation/poets/run.sh
```diff
@@ -0,0 +1,5 @@
+for d in scripts/* ; do
+    echo "$d"
+    /pash/pa.sh $d > /dev/null
```
angelhof, line 3: I would replace `/pash` with `$PASH_TOP` so that this works for machines where pash is not under `/pash`

evaluation/poets/NOTES
```diff
@@ -0,0 +1,24 @@
...
+        scripts/vowel_sequencies_gr_1K.sh
+
+    -> trigram script does not exist
+        scripts/trigram_rec.sh
+
+Several scripts had different outputs in comparison with bash:
+    eg:
+        scripts/4letter_words.sh 
+    pash output: 
+        4767
```
angelhof, line 17: The issue seems to be grep -c. The way to address this would be to modify grep's annotation to consider `grep -c` to be pure instead of `stateless`.

evaluation/poets/scripts/1syllable_words.sh
```diff
@@ -0,0 +1,3 @@
+# 1-syllable words 
+INPUT=${INPUT:-inputs/genesis}
```
angelhof, line 2: Similarly, I would replace `inputs/genesis` with the path relative to `$PASH_TOP`.

angelhof (changes requested):
Thanks a lot :) I have left some comments. If you need help with modifying the annotation of grep, let me know.

evaluation/poets/scripts/1syllable_words.sh
nvasilakis: (Update for @angelhof : we are moving inputs to point to `$PASH_TOP/evaluation/scripts/input/` and updating `gen.sh` to pull the right inputs for these scripts.)

evaluation/poets/NOTES
nvasilakis: Adding @gyalpodongo here, who might be able to help with this.

nvasilakis (approved):
@dkarnikis the only thing missing is updating `fetch.sh` to fetch input files. Other than that, looks good!

evaluation/poets/scripts/1syllable_words.sh
```diff
@@ -0,0 +1,3 @@
+# 1-syllable words 
+INPUT=${INPUT:-$PATH_TOP/evaluation/scripts/input/genesis}
```
angelhof, line 2: Please replace `$PATH_TOP` with `$PASH_TOP`

nvasilakis: Wow, great catch!

evaluation/poets/scripts/verses_2om_3om_2instances.sh
```diff
@@ -1,5 +1,5 @@
 # verses with 2 or more, 3 or more, exactly 2 instances of light.
-INPUT=${INPUT:-$PATH_TOP/evaluation/scripts/input/genesis}
-grep -c 'light.*light' ${INPUT} 
-grep -c 'light.*light.*light' ${INPUT}
-grep 'light.*light' ${INPUT} | grep -vc 'light.*light.*light' 
+INPUT=${INPUT:-$PASH_TOP/evaluation/scripts/input/poets/genesis}
+grep -c 'light.*light' ${INPUT} | paste -sd+ | bc
```
angelhof, line 3: If possible, let's keep the original version of the scripts and put the alternative version (that doesn't have the issue) in a comment. The issue that happens now is solvable easily by modifying the grep annotation.

angelhof: Commenting out the `paste -sd+ | bc` should ideally happen for all `grep -c` occurences.

dkarnikis:
Reverted the scripts to match the originals. The problem with awk and grep -c remains. scripts/trigrams_rec.sh invokes a second script, can the script be hardcoded into trigrams_rec?

angelhof (approved):

angelhof:
Thanks!

> scripts/trigrams_rec.sh invokes a second script, can the script be hardcoded into trigrams_rec?

For now we could leave it as is, and we can see what to do with it when we run a proper evaluation of PaSh on these scripts.

nvasilakis:
@angelhof We're inlining this script for now, and correcting double quotes to single quotes for `awk`.

nvasilakis (approved):
Awesome, thanks!

PR #132, opened 2021-02-16
Modified the dataset. Works only on bash

dkarnikis:


evaluation/aliases/run_alias.sh
```diff
@@ -0,0 +1,25 @@
...
+cd $PASH_TOP/evaluation/scripts/input/
+# we could read the file iteratively with IFS, but the environment was affected
+IFS=$'\r\n' GLOBIGNORE='*' command eval  'cmd_array=($(cat generated.file))'
+lc=$(cat generated.file | wc -l)
+for i in $(seq 0 $lc)
+do
+    # get the entry from the array
+    p=${cmd_array[$i]}
+    # add a timeout to our script
+    timeout --signal=SIGINT 50s /bin/bash -e $p >> /dev/null 2>&1  #./cmd.sh #eval "bash ./cmd.sh"
```
nvasilakis, line 13: The timeout was a great idea.

I am wondering if `$p` causes problems due to expansion (i.e., interpreted as bash arguments rather than a command). One idea to avoid this would be to put each line into a file and run it.

```sh
head -n $i $INPUT | tail -1 > cmd.sh
./cmd.sh
```

nvasilakis (approved):
I'm approving and merging, but please remember to run this.

PR #137, opened 2021-02-17
Infra for checking that aggregators work as expected

nvasilakis:
This builds a basic infrastructure for testing annotations and aggregators.

### DIFF
--- a/annotations/sort.json
+++ b/annotations/sort.json
@@ -1,5 +1,6 @@
 {
     "command": "sort",
+    "aggregate": "sort -m",
     "cases":
     [
         {
@@ -15,6 +16,7 @@
         {
             "predicate": "default",
             "class": "pure",
+            "agg": "sort",
             "inputs": ["stdin"],
             "outputs": ["stdout"]
         }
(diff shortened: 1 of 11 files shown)


annotations/sort.json
```diff
@@ -15,6 +16,7 @@
         {
             "predicate": "default",
             "class": "pure",
+            "agg": "sort",
```
angelhof, line 19: I am not sure I understand why there are two aggregate fields `aggregate` at the top and `agg` in the predicate. The aggregator should only be in the predicate since it might only work for a specific invocation. We can merge for now but this should be fixed :)

annotations/tests/test-core.sh
```diff
@@ -0,0 +1,26 @@
...
+    echo "Must set PASH_TOP" 1>&2
+    exit 1
+fi
+
+IN=$PASH_TOP/evaluation/scripts/input/1M.txt
+
+CMD="$1"
+FLG="$2"
+# AGG=$(grep aggregate ../$CMD.json | cut -d ':' -f '2' | awk '{$1=$1};1' | tr '"' ' ' | sed 's/,$//')
+AGG=$(cat ../$CMD.json | jq -r '.aggregate' | tr -d '$' | sed "s;PASH_TOP;$PASH_TOP;")
```
angelhof, line 16: I guess that is where you use the top level `aggregate`. I guess it is fine for now.

angelhof (approved):

PR #138, opened 2021-02-17
Add a few DGSH scripts

nvasilakis:
This is a work-in-progress port.


nvasilakis:
Could we maybe name setup files as `n-setup.sh`, where _n_ is the script number?

PR #142, opened 2021-02-19
Add Python/C libdash port and change config.py to use the C port

thurstond:
The C port passes test_evaluation_scripts.sh. I haven't plugged in the Python scripts, but I'm including them as well because it's too messy to extract just the C files.

N.B. the C port requires JSON-C, which is available separately (I don't know how to link the repo or add it to the pash docker image):

```
wget https://s3.amazonaws.com/json-c_releases/releases/json-c-0.15.tar.gz
tar zxf json-c-0.15.tar.gz
cd json-c-0.15
mkdir build
cd build/
cmake -DCMAKE_INSTALL_PREFIX=/pash/json-c-0.15/install ../
make install
```


angelhof:
Question: Does the Python version also need JSON-C?

thurstond:
The Python version doesn't need JSON-C.

angelhof:
Then maybe we should just wait and merge the python version (to not even have to deal with the JSON-C dependency)? Or is that not possible yet?

thurstond:
I thought we wanted to do a performance comparison with the C version as well?

The Python functions for Pash AST <-> shell script (as a Python string) are there, but the existing functions in parse.py assume the shell scripts are stored in files on disk. It'd be easy enough to write a Python script wrapper that does that, but we'd lose the performance benefits of not spawning processes. Hence, I'd need to look deeper and refactor the Pash code to use Python strings instead of files (where possible).

angelhof:
> I thought we wanted to do a performance comparison with the C version as well?

The performance comparison is not on the critical path I think. We could do it later anyway right?

> The Python functions for Pash AST <-> shell script (as a Python string) are there, but the existing functions in parse.py assume the shell scripts are stored in files on disk. It'd be easy enough to write a Python script wrapper that does that, but we'd lose the performance benefits of not spawning processes. Hence, I'd need to look deeper and refactor the Pash code to use Python strings instead of files (where possible).

If you can just add a wrapper for now that would be great. The refactoring for additional performance can happen on another PR (someone else could also do it) :)

nvasilakis:
I believe we (I?) had implemented a wrapper about a year ago, I'll check to see if it's anywhere.

nvasilakis:
...just created a gist, in case it's useful: https://gist.github.com/nvasilakis/40d17648d864cbe8f2438029691f3842

thurstond:
> ...just created a gist, in case it's useful: https://gist.github.com/nvasilakis/40d17648d864cbe8f2438029691f3842

Those wrappers are in the reverse direction e.g., they take a (stdin -> stdout) program, and convert it to (Python string <-> Python AST).
I have the the (Python string <-> Python AST) functions; parse.py wants a (stdin -> stdout) program.

I suspect the wrappers you wrote imply that the reverse wrapper I need actually cancels out and we actually need neither of them (at the cost of some refactoring).

thurstond:
What I wrote above isn't quite true; I do have shell script -> AST (parse_to_ast2.py). Anyway, I'll write the (reverse) wrappers tomorrow and submit a new pull request with them.

nvasilakis:
(You can keep pushing to the same remote branch while the PR is open:-)

PR #139, opened 2021-02-18
Remove now superfluous remote/

zyrolasting:
`pash-meta` now has a more up to date copy of the `remote` directory. Keeping this one around will only create confusion.

Do not merge until `ci-perf.sh` can operate without Node.js.

PR #144, opened 2021-02-19
Use only Bash when summarizing performance test results

zyrolasting:
Make `ci-perf.sh` use only Bash. I also set `ci-perf.sh` so that you can either execute it, or source it. That proves helpful when one part of the script doesn't work and you need to iterate on that part.

Note that I verified this without running performance tests. I did this by iterating only on the `summarize_perf_suite` function. I don't expect this to be a problem because that's the only part of the code being replaced from the test runner's perspective.

Unblocks #139

### DIFF
--- a/scripts/ci-perf.sh
+++ b/scripts/ci-perf.sh
@@ -34,42 +34,15 @@ main() {
     # used to run the tests.
     git checkout "$latest_main_revision";
 
-    summarize_suite() {
-        local heading="$1";
-        local summary_name="$2";
-        local subdir="$3";
-        local tests="$4";
-        local width="$5";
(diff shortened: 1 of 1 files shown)


nvasilakis:
This looks good. Was that the only use of node for summarization? i.e., can we remove the node dependency from the docker file?

zyrolasting:
@nvasilakis Yes. See #139

angelhof (approved):

nvasilakis (approved):
OK, let's merge and we can test!

nvasilakis:
Instead of parsing all these times and converting to seconds, it would make sense to just have the scripts output in seconds.

PR #145, opened 2021-02-19
Add Python/C libdash port and change config.py to use the Python port

thurstond:
This changes Pash to use the Python port of libdash instead of OCaml.

- You may need to set parse_to_json2.py and json_to_shell2.py to be executable.
- Both the Python and C ports pass test_evaluation_scripts.sh, though the C port requires JSON-C and is not enabled by default.

Technical notes (code changes since yesterday):
- Added wrappers for shell script <-> JSON (parse_to_json2.py, json_to_shell2.py). Slower than skipping JSON, but makes it a drop-in replacement for the OCaml binaries.
- I've modified the Python port so that it doesn't use the dash2.c shims from the C port, so that there's no need to compile anything. Relatedly: refactored the distribution of code in parse_to_ast2.py vs dash2.py, making it more similar to the OCaml version.
- Improved ast2shell's compatibility with JSON formats (somewhere in Pash outputs booleans as text e.g., "false", which evaluates to True in Python)


nvasilakis:
Epic PR.

nvasilakis:
(Also, love the commit message`:wq!ZZ`)

angelhof:
Thanks a lot Thurston, I will start playing with this PR, fixing some absolute paths that only work in docker, merging with main, and then testing and merging it.

angelhof:
I am commenting that here so that we don't forget. Let's make this PR on a `future`/`dev` branch instead of `main` so that main stays frozen.

angelhof:
Moving this to #147

PR #146, opened 2021-02-20
Fix on pash-c implementation

dkarnikis:
The current implementation uses a fixed name (pash_immediate_command.sh) to write the input for the command to execute. This might be a problem when executing multiple 'pash -c' commands on the same path as all of them will write on the same file?

### DIFF
--- a/compiler/pash.py
+++ b/compiler/pash.py
@@ -17,20 +17,7 @@ import shutil
 def main():
     preprocessing_start_time = datetime.now()
     ## Parse arguments
-    args = parse_args()
-    config.pash_args = args
-
-    ## Initialize the log file
-    config.init_log_file()
-    if not config.config:
-        config.load_config(args.config_path)
-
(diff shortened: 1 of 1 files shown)


nvasilakis (approved):
Has been tested and looks good.

compiler/pash.py
```diff
@@ -61,14 +48,16 @@ def main():
...
-        execute_script(fname, args.debug)
+        execute_script(fname, args.debug, args.command)
 
 
 def parse_args():
     prog_name = sys.argv[0]
     if 'PASH_FROM_SH' in os.environ:
         prog_name = os.environ['PASH_FROM_SH']
     parser = argparse.ArgumentParser(prog_name)
+    # group = parser.add_mutually_exclusive_group(required=True)
```
angelhof, line 59: If this is not used let's delete it before merging :)

angelhof (approved):

PR #153, opened 2021-02-23
Aliases WIP

dkarnikis:


angelhof:
Could you redirect this PR to `future` instead of main? We are keeping `main` frozen for the next couple weeks.

PR #155, opened 2021-02-23
Produce a tex table for eurosys small one-liners results

angelhof:


zyrolasting:
I'm not quite sure from first glance, but it looks like this change might break https://github.com/andromeda/pash/blob/main/scripts/ci-perf.sh#L39

Can you verify?

angelhof:
It should't break anything since it just modifies our previous plotting infrastructure (which is not run anywhere in `ci-perf` if I am not mistaken)

nvasilakis (approved):

PR #156, opened 2021-02-23
Add Dockerfile for Smoosh tests

zyrolasting:
Add Pash's testing facilities to an existing Pash image using separate Dockerfile.

I'm torn between doing this and just adding to the first Dockerfile. I went for this approach because we discussed multiple Dockerfiles over email. Composing `ARG` and `FROM` is the simplest option I've seen for reusing code for now, but I think it would be hard to understand if we ever use many more Dockerfiles.


zyrolasting:
Correctness tests run to completion on the CI worker with the updated image. But before we merge, is this a problem?

```
./shell_tests.sh: 2: /smoosh/tests/default.env: --width: not found
```

angelhof:
This warning is not a problem, I think it happens when the smoosh testing infrastructure tries to source pash. We can ignore it.

angelhof (approved):

PR #170, opened 2021-03-23
Expansion in Python, without calling `bash`

mgree:
Use our own expansion rather than deferring to bash.

There are a number of shortcuts:

 - Unimplemented features (arithmetic, pattern matching, globbing)
 - Imperfect field splitting
 - Incomplete handling of special variables (`$@`, `$*`, positionals, `$?`, etc.)


compiler/ast_to_ir.py
```diff
@@ -106,8 +107,16 @@ def compile_asts(ast_objects, fileIdGen, config):
         # log("Compiling AST {}".format(i))
         # log(ast_object)
 
+        log("MMG config")
+        log(config['shell_variables'])
```
angelhof, line 111: If these are not needed we could remove them since the log would fill up with these calls to `ast_to_shell`. Maybe we could use the `level` argument of the `log` function to make them have a high debug level so that they do not appear when using `-d 1` ?

compiler/config.py
```diff
@@ -15,13 +15,16 @@
...
+PRINTER_BINARY = os.path.join(PASH_TOP, "compiler/parser/json_to_shell.native")
+
 PYTHON_VERSION = "python3"
 PLANNER_EXECUTABLE = os.path.join(PASH_TOP, "compiler/pash_runtime.py")
 RUNTIME_EXECUTABLE = os.path.join(PASH_TOP, "compiler/pash_runtime.sh")
 
 ## This is set in pash.py and pash_runtime.py accordingly.
 ## In both cases the setting is different.
-PASH_TMP_PREFIX = None
+PASH_TMP_PREFIX = "pash_"
```
angelhof, line 27: Is this necessary here? I think that it is set properly in `pash.py` and `pash_runtime.py`.

compiler/config.py
```diff
@@ -15,13 +15,16 @@
 else:
     PASH_TOP = subprocess.run(GIT_TOP_CMD, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True).stdout.rstrip()
 
+PARSER_BINARY = os.path.join(PASH_TOP, "compiler/parser/parse_to_json.native")
+PRINTER_BINARY = os.path.join(PASH_TOP, "compiler/parser/json_to_shell.native")
```
angelhof, line 19: Similarly, are those needed? We are now using Thurston's infrastructure so parsing and printing should not happen through ocaml

compiler/config.py
```diff
@@ -144,13 +147,16 @@ def init_log_file():
...
+
     config['shell_variables'] = None
     config['shell_variables_file_path'] = var_file_path
     if(not var_file_path is None):
         vars_dict = {}
         with open(var_file_path) as f:
             lines = [line.rstrip() for line in f.readlines()]
 
+        # ??? MMG 2020-12-10 not sure how you're generating this file, but what about variables with newlines in them?
+        # MMG 2021-03-09 definitively breaking on IFS and function outputs (i.e., `declare -f`)
```
angelhof, line 159: Yes this is indeed very hacky, and we will probably need to fix this for smoosh tests to pass.

compiler/ast_to_ir.py
```diff
@@ -378,7 +389,7 @@ def execute_shell_asts(asts):
     save_asts_json(asts, ir_filename)
     output_script = from_ir_to_shell(ir_filename)
     # log(output_script)
-    exec_obj = subprocess.run(["/usr/bin/env", "bash"], input=output_script,
+    exec_obj = subprocess.run(["/bin/bash"], input=output_script,
```
angelhof, line 392: A user had an issue with this. Did you change it back on purpose? If so we should find a solution that works for everyone.

compiler/config.py
```diff
@@ -159,15 +165,35 @@ def read_vars_file(var_file_path):
...
+            else:
+                var_name = rest[:eq_index]
+                var_value = rest[(eq_index+1):]
+
+            ## Strip quotes
+            if var_value is not None and len(var_value) >= 2 and \
+               var_value[0] == "\"" and var_value[-1] == "\"":
+                var_value = var_value[1:-1]                
+                
+            print("MMG writing var {}={} with type {}".format(var_name,var_value, var_type))
```
angelhof, line 196: Let's replace this `print` with a `log`

angelhof (changes requested):
I have left some comments for changes, but for most of those I can also take care of them if you agree :)

mgree:
Went ahead and made these fixes, which are still passing on my end. I'm still a wee bit confused about the Docker setup, where I'm needing to copy in files from `compiler/parser/ceda` to get things to go... CI will figure out the ground truth here, right? (Sorry to be dense.)

angelhof (approved):

angelhof (approved):

angelhof:
@nvasilakis if it looks OK let's merge

PR #169, opened 2021-03-17
Round split final (and wrap, merge, unwrap, etc)

tammam1998:
Fixed some history issues with the previous pull request. Contains a working implementation of r_split, r_wrap, r_unwrap, r_merge tested with microbenchmarks.  Also contains some log parsing, testing, and plotting utilities used in benchmarking and comparison to normal auto-split.

PR #171, opened 2021-03-23
In memory buffering

tammam1998:
Implements memory buffering using dgsh-tee. Replaces all eager calls with dgsh-tee calls when --dgsh_tee flag is passed.


tammam1998:
depends on #169

runtime/r_wrap.c
```diff
@@ -135,17 +135,17 @@ void processCmd(char *args[])
                 // Write to forked process
                 // For some reason nonblocking fd is not working correctly with fwrite. 
                 // Check shortest-scripts.sh with big batch size to recreate the problem
-                if ((len = write(outputFd, writebuffer, currWriteLen)) < 0) {
-					switch (len) {
+                if ((len = write(outputFd, writebuffer, currWriteLen)) > 0) {
+                    currWriteLen -= len; 
+                    memmove(writebuffer, writebuffer + len, currWriteLen);
```
angelhof, line 140: What is this change for?

compiler/test_evaluation_scripts.sh
```diff
@@ -23,6 +23,7 @@ n_inputs=(
 configurations=(
     ""
     "--r_split"
+    "--dgsh_tee"
```
angelhof, line 26: Maybe we want to have a combination of r_split and dgsh tee (since that is the optimal one that we really care about anyway.

angelhof (approved):
Let's wait before merging to check if all tests pass after the evaluation restructuring and fixes.

compiler/test_evaluation_scripts.sh
tammam1998: Added it!

runtime/r_wrap.c
tammam1998: I think this was a small change for minor performance improvement. The processor usually speculates the true branch faster than the other one but it's not really doing much here.

nvasilakis:
@angelhof This can be tested now.

compiler/config.py
```diff
@@ -90,6 +90,9 @@ def add_common_arguments(parser):
                         type=int,
                         help="(experimental) configure the batch size of r_splti (default: 100KB)",
                         default=100000)
+    parser.add_argument("--dgsh_tee",
+                        help="(experimental) use dgsh-tee instead of eagers",
```
nvasilakis, line 94: either `dgsh-tees` or `eager`:-)

evaluation/dgsh-tee/manual-tests/shortest-scripts.sh
```diff
@@ -0,0 +1,221 @@
...
+mkfifo "#file60"
+mkfifo "#file61"
+mkfifo "#file62"
+
+mkfifo "#file63"
+mkfifo "#file64"
+mkfifo "#file65"
+mkfifo "#file66"
+
+{ cat /home/tammam/pash/evaluation/scripts/input/1G.txt >"#file2" & }
```
nvasilakis, line 113: Could we please use `$PASH_TOP` here? We won't be able to run these tests on other machines otherwise:-)

nvasilakis: (and below)

runtime/Makefile
```diff
@@ -40,5 +49,20 @@ r-wrap: r_wrap.c r_split.h
 
 r-unwrap: r_unwrap.c r_split.h
 	gcc ${CFLAGS} r_unwrap.c -o r_unwrap
+
+
+libdgsh_a_SOURCES = negotiate.c $(DGSH_ASSEMBLY_FILE)
+dgsh_tee_SOURCES = dgsh-tee.c
+
+dgsh-tee:
+	git clone https://github.com/tammam1998/dgsh.git;\
```
nvasilakis, line 58: Would it make sense to fold this dependency under pash/runtime? (cc @angelhof )

nvasilakis (approved):
Left some comments, looks great!:-)

runtime/r_wrap.c
angelhof: I would be really surprised if that was true (speculating faster due to such a code change) :)

runtime/Makefile
angelhof: Do you mean using a submodule? The reason why @tammam1998 did it like this (correct me if I am wrong) is that dgsh tee is a very small part of dgsh and so this way it is just cloned, and then everything that is not necessary is deleted.

angelhof:
Testing

runtime/Makefile
nvasilakis: sounds good, thanks!

PR #179, opened 2021-03-29
Fix tests after restructuring of evaluation

angelhof:
This PR fixes the tests, making sure that they can run under the new evaluation structure.

I included words, all_cmds, and 1M.txt so that the tests can run without anything being online.


angelhof:
@nvasilakis if you find that OK, feel free to merge

angelhof:
We need to figure out what is wrong with the link, because it now returns not found

angelhof:
@nvasilakis both ` evaluation/tests/input/all_cmds.txt` and ` evaluation/tests/input/words` also need to be pulled from somewhere (here and in the `evaluation/benchmarks/oneliners/input/setup.sh` where they are currently not pulled).

nvasilakis:
@angelhof it turns out that the URL prefix was wrong — not sure how we missed this after that time when `apha.ndr.md` went offline a year ago (I can now see why "this hasn't worked more times than it has":-). I added generation fallback when links fail, and we can copy to other input scripts (cc @dkarnikis)

When ready, please merge-by-squash because to jettison all obsolete input files:-)

compiler/test_evaluation_scripts.sh
```diff
@@ -5,16 +5,21 @@ export PASH_TOP=${PASH_TOP:-$(git rev-parse --show-toplevel --show-superproject-
...
 
 echo "Deleting eager intermediate files..."
 rm -f /tmp/eager*
 mkdir -p $intermediary_dir
 rm -rf "$test_results_dir"
 mkdir -p "$test_results_dir"
 
+echo "Generating inputs..."
+cd "$microbenchmarks_dir/input"
+./setup.sh
```
angelhof, line 20: @nvasilakis so I added this here to make sure that inputs are generated before running tests, but we wouldn't want to re-download inputs whenever we run tests. How should we improve that? Should we check if the files exist before running the setup (maybe having a check-setup) file or should we just run the setup for tests in the installation script?

nvasilakis: That's a great point, and we do some of that in other `setup.sh` scripts. Probably the best way to handle this would be in `setup.sh` script itself (cc @dkarnikis). This way every `setup.sh` would be self-contained, and we don't have to remember to call more scripts or check values. I just updated `setup.sh` to handle this. What do you think?

nvasilakis (approved):
Looks good:-)

PR #185, opened 2021-04-03
Fixes in evaluation benchmarks

angelhof:
I am opening this as a draft to keep track of all the changes that need to be done to make the new benchmarks work, with and without pash. I will fix some of the issues myself, but let's discuss the rest.


angelhof:
@nvasilakis Can we abstract the functionality for input generation? 

We want to support removing the inputs with `-c`, not redownloading/regenerating them if they exist, and always adding a newline to them in the end if they dont already have one (`scripts/append_nl_if_not.sh`). 

Maybe we can make a function `generate_input` that takes an file name and a script that generates it to stdout, and then checks if it exists, if not it generates it using the script, and then appends a newline if needed in the end.

angelhof:
It would make sense to have some form of progress output in setup.sh scripts (especially ones that download GB of data).

nvasilakis:
I'm noting here so that I don't forget:
1. We can easily have different inputs for each one of oneliners — that's the point of the entire `run.seq.sh`, to orchestrate execution of the set as needed. Can we add what's needed in the script in terms of `$IN`?
2. Truly abstracting input generation is tricky because different inputs have different ways the can be generated if the URL doesn't work. We'd want to keep "email pash authors" as a last resort—we can basically generate all inputs if we want. 
3. Progress is easy, we can just remove -s from curl
4. `cd $benchmark_set` should come before each check to inputs — we'd like to have output per benchmark set in each benchmark set itself, rather than outside.
5. variables like `IN_PRE` have to be set outside otherwise the script is going to inherit whatever exist in the environment. We can also set them in individual benchmarks, so that they can be called standalone (say, from Python scripts, or individually)

evaluation/benchmarks/unix50/input/setup.sh
```diff
@@ -2,9 +2,32 @@
...
+
+for input in ${inputs[@]}
+do
+  if [ ! -f "${input}.txt" ]; then
+    wget "http://ndr.md/data/unix50/${input}.txt"
+    "$PASH_TOP/scripts/append_nl_if_not.sh" "${input}.txt"
+  fi
+done
+
+## FIXME: Calling this script with --full is not idempotent.
```
nvasilakis, line 30: There's a more important problem here, that we don't have a way (yet) to get the original small inputs from a source that's not one of our servers. How can we fix that? (if we had that, then everything's solved).

evaluation/benchmarks/run.seq.sh
```diff
@@ -2,7 +2,8 @@
 
 # FIXME: skip running if output file exists (using tee?)
 
-set -e
+## FIX: We should not have a set -e in a script that is supposed to be sourced.
+# set -e
```
nvasilakis, line 6: It was helpful while writing the script, you can delete it.

evaluation/benchmarks/oneliners/input/setup.sh
```diff
@@ -5,27 +5,101 @@ set -e
...
-  touch 3G.txt
-  for (( i = 0; i < 3; i++ )); do
-    cat 1G.txt >> 3G.txt
-  done
 
-  touch 10G.txt
+if [ ! -f ./1M.txt ]; then
+  curl -sf 'http://ndr.md/data/dummy/1M.txt' > 1M.txt
+  if [ $? -ne 0 ]; then
+    echo 'cannot find 1M.txt -- please contact the developers of pash'
```
nvasilakis, line 24: These we can recreate easily — see the setup of `tests`

evaluation/benchmarks/oneliners/input/setup.sh
```diff
@@ -5,27 +5,101 @@ set -e
...
   for (( i = 0; i < 10; i++ )); do
-    cat 10G.txt >> 100G.txt
+    cat 10M.txt >> 100M.txt
   done
 fi
+
+if [ ! -f ./1G.txt ]; then
+  curl -sf 'http://ndr.md/data/dummy/1G.txt' > 1G.txt
+  if [ $? -ne 0 ]; then
+    echo 'cannot find 1G.txt -- please contact the developers of pash'
```
nvasilakis, line 47: These we can recreate easily — see the setup of `tests`

evaluation/benchmarks/oneliners/input/setup.sh
```diff
@@ -5,27 +5,101 @@ set -e
...
+    exit 1
+  fi
+  "$PASH_TOP/scripts/append_nl_if_not.sh" ./1G.txt
+fi
+
+# download wamerican-insane dictionary and sort according to machine
+if [ ! -f ./dict.txt ]; then
+  curl -sf 'http://ndr.md/data/dummy/dict.txt' | sort > dict.txt
+  if [ $? -ne 0 ]; then
+    echo 'cannot find dict.txt -- please contact the developers of pash'
```
nvasilakis, line 57: These we can recreate easily — see the setup of `tests`

evaluation/benchmarks/oneliners/input/setup.sh
```diff
@@ -5,27 +5,101 @@ set -e
...
-curl 'http://ndr.md/data/dummy/dict.txt' | sort > dict.txt
+PASH_TOP=${PASH_TOP:-$(git rev-parse --show-toplevel)}
 
-if [ "$#" -eq 1 ] && [ "$1" = "--full" ]; then
-  echo Generting full-size inputs
-  # FIXME PR: Do we need all of them?
+# another solution for capturing HTTP status code
+# https://superuser.com/a/590170
+
+input_files="1M.txt 10M.txt 100M.txt 1G.txt dict.txt 3G.txt 10G.txt 100G.txt all_cmds.txt all_cmdsx100.txt"
```
nvasilakis, line 13: Up to 1G everything is on the server. I think I'll take care of fixing this.

nvasilakis:
I left a bunch of comments. I will try to fix some of them later today. It generally looks like many of these could be easily abstracted away, including going through tests in a directory.

evaluation/benchmarks/run.seq.sh
```diff
@@ -13,94 +14,104 @@ if [[ -z "$PASH_TOP" ]]; then
...
-  echo 30.sh: $({ time ./30.sh > /dev/null; } 2>&1) | tee -a ./seq.res
-  echo 31.sh: $({ time ./31.sh > /dev/null; } 2>&1) | tee -a ./seq.res
-  echo 32.sh: $({ time ./32.sh > /dev/null; } 2>&1) | tee -a ./seq.res
-  echo 33.sh: $({ time ./33.sh > /dev/null; } 2>&1) | tee -a ./seq.res
-  echo 34.sh: $({ time ./34.sh > /dev/null; } 2>&1) | tee -a ./seq.res
-  echo 35.sh: $({ time ./35.sh > /dev/null; } 2>&1) | tee -a ./seq.res
-  echo 36.sh: $({ time ./36.sh > /dev/null; } 2>&1) | tee -a ./seq.res
+  export IN_PRE=$PASH_TOP/evaluation/benchmarks/unix50/input
+
+  for number in `seq 36`
```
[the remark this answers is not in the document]
angelhof: We can abstract the seq 36 to a variable (then we can grep it or do whatever for the loop range). Would that work? Copying them leads to a ton of duplication on this and the pash file.

evaluation/benchmarks/run.seq.sh
```diff
@@ -13,94 +14,104 @@ if [[ -z "$PASH_TOP" ]]; then
...
 }
 
 web-index(){
-  cd web-index/
   if [ -e ./seq.res ]; then
     echo "skipping $(basename $(pwd))/seq.res"
     return 0
   fi
 
+  cd web-index/
```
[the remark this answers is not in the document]
angelhof: I haven't had time to make that work yet, I will get to it at some point today/tomorrow

nvasilakis:
Merging this for now and can keep adding fixes.

PR #189, opened 2021-04-05
fixed bug in r_split

tammam1998:


runtime/r_wrap.c
```diff
@@ -124,7 +124,7 @@ void processCmd(char *args[])
 					        case EAGAIN:
                                 break;
                             default:
-                                err(2, "r_wrap: failed reading from fork, error %d", errno);
+                                err(2, "Failed reading from fork, error %d", errno);
```
angelhof, line 127: I think it was useful seeing that r_wrap was the issue for these errors. Did you remove that for a specific purpose?

tammam1998: If you noticed it was duplicated, the err function prefix the message with "<filename>: " so no need for it.

angelhof: Oh thanks!

angelhof (approved):

PR #196, opened 2021-04-07
Quick abort

angelhof:
Mostly working, however if the sequential is done first, the compiler's log is produced in a delayed fashion, having a little clunky behavior when running in an interactive shell with `-d 1` and without a log_file.

It is not thoroughly tested yet, but we should merge as a first step.


nvasilakis (approved):
Looks great!!

PR #198, opened 2021-04-07
Posh Fixes

dkarnikis:


evaluation/benchmarks/run.seq.sh
```diff
@@ -353,7 +355,7 @@ posh() {
 
 
 
-#posh
+posh
```
nvasilakis, line 358: Maybe comment out because this script is to be sourced

nvasilakis (approved):

PR #201, opened 2021-04-09
Add MTS analytics in top evaluation script

angelhof:


nvasilakis (approved):
Note to self: some of these variables/identical lines out of these functions; possibly try to make more lines identical and pull them out.

PR #202, opened 2021-04-09
Improve logging and cleanup for quick-abort runtime

angelhof:


angelhof:
Merging so that I can keep exploring a bug

PR #204, opened 2021-04-12
Fix for fork running out of resources

tammam1998:


runtime/r_wrap.c
```diff
@@ -180,14 +184,17 @@ void processCmd(char *args[])
...
+            // write block to stdout
             writeHeader(stdout, id, currReadLen);
             safeWriteWithFlush(stdoutBlock, 1, currReadLen, stdout);
 
-            //update header (ordered at the end so !feof works) and cleanup
+            // update header (ordered at the end so !feof works) and cleanup
             readHeader(stdin, &id, &blockSize);
+
+            // wait is necessary to reclaim process
+            // Do I need to do something with return status?
```
angelhof, line 195: I think the return status of the parallel does not correspond to the one from the sequential so it can be ignored.

angelhof (approved):
Looks good! @tammam1998 did the tests pass with this?

PR #205, opened 2021-04-12
Added bash -c for every r_wrap

tammam1998:
This is a fix to allow r_wrap to execute bash functions. Functions still need to be exported for r_wrap to be able to use them. A better long-term solution would be to only add bash -c when calling r_wrap with a function.

### DIFF
--- a/annotations/custom_tr.json
+++ b/annotations/custom_tr.json
@@ -1,0 +1,12 @@
+{
+    "command": "custom_tr",
+    "cases":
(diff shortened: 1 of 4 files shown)


nvasilakis (approved):

angelhof (changes requested):
Looks good, but I will add a test to make sure that there is no regression.

angelhof (approved):

PR #206, opened 2021-04-12
Improve setup scripts

nvasilakis:
This improves setup scripts:
* removes experiment dependencies from main installation path
* decouples distro-dependencies (that may require `sudo` and are most useful in container environments) from project setup
* adds distro-dependencies for other distributions
* removes the need for `sudo` in project setup
* improves on several setup files and experimental infrastructure


compiler/parser/Makefile
```diff
@@ -38,7 +38,7 @@ clean :
 	$(JSON_TO_SHELL).native $(JSON_TO_SHELL).byte test.err
 
 libdash:
-	cd libdash && ./autogen.sh || ./autogen.sh && ./configure && make && sudo make install
+	cd libdash && ./autogen.sh && ./configure && make
```
angelhof, line 41: The reason why I had this is that sometimes it failed spuriously on the first run (I have to check again on my machine whether this happens). We can merge for now and remember this if an issue appears.

angelhof (approved):
Looks good to me, as long as it works on a new machine :)

compiler/parser/Makefile
nvasilakis: Noted!

PR #209, opened 2021-04-12
Dgsh and run.seq fixes

dkarnikis:


evaluation/benchmarks/run.seq.sh
```diff
@@ -369,7 +398,7 @@ posh() {
 
 
 
-posh
+#posh
 #poets
 #aliases
 dgsh
```
nvasilakis, line 404: comment out dgsh too:-)

nvasilakis (approved):

PR #216, opened 2021-04-13
Fixes on tutorial scripts

dkarnikis:


nvasilakis (approved):
Noting here so that we don't forget:
1. revert `dirname $0` and open issue
2. revert `up.sh` and add note in README.md saying to run `./scripts/distro-deps.sh; ./scripts/setup-pash.sh` manually (if they want to inspect the scripts).

scripts/install.sh
```diff
@@ -22,7 +22,7 @@ shift "$(( OPTIND - 1 ))"
 
 ## If option -p is set, also run the sudo
 if [ "$prepare_sudo_install_flag" -eq 1 ]; then
-  ./distro-deps.sh
+  ./scripts/distro-deps.sh
```
nvasilakis, line 25: `scripts/install.sh` is already in `scripts`, why do we need to change this?

scripts/install.sh
```diff
@@ -38,4 +38,4 @@ else
   done
 fi
 
-./setup-pash.sh
\ No newline at end of file
+./scripts/setup-pash.sh
```
nvasilakis, line 41: same here

nvasilakis: Ah, there's a `cd $PASH_TOP`, can we remove it?

nvasilakis (approved):

nvasilakis (approved):
Looks great, thanks!

PR #219, opened 2021-04-13
dgsh fixes on setup and run.seq

dkarnikis:


evaluation/benchmarks/run.seq.sh
```diff
@@ -463,3 +463,5 @@ posh() {
     done
     cd ..
 }
+
+dgsh
```
nvasilakis, line 467: comment out?

dkarnikis: Fixed

nvasilakis (approved):
Good catch:-)

PR #220, opened 2021-04-13
Handle non-utf8 characters in spell.sh

dkarnikis:


nvasilakis (approved):
Thanks! (cc @angelhof: `col` was failing on some non-utf8 inputs, so @dkarnikis added an additional pipepline stage. We don't think we need to add it to `demo-spell.sh` because it adds a dependency (`iconv`), which is a relatively important risk for a starting example that is supposed to work everywhere. )

Issue #218, opened 2021-04-13
dirname $0 evaluates to /tmp


nvasilakis:
It looks like PaSh evaluates `dirname $0` to `/tmp` in [evaluation/intro/demo-spell.sh](./evaluation/intro/demo-spell.sh).

angelhof:
Addressed by #246

PR #223, opened 2021-04-13
Quickfix demo-spell

nvasilakis:
Quick fix for #218

### DIFF
--- a/evaluation/intro/demo-spell.sh
+++ b/evaluation/intro/demo-spell.sh
@@ -1,10 +1,14 @@
-#!/bin/bash
+#!/bin/sh
 
-# Would this expansion work:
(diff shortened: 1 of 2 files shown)


evaluation/intro/demo-spell.sh
```diff
@@ -1,10 +1,14 @@
-#!/bin/bash
+#!/bin/sh
```
angelhof, line 1: Why is this change necessary?

angelhof (approved):

evaluation/intro/demo-spell.sh
nvasilakis: It's not necessary, just want to make sure this example runs with `sh` (and it does)

PR #228, opened 2021-04-19
Improved r split

tammam1998:
- Improved bar plotting script
- Improved memory usage when the file doesn't contain newlines or very few. The header now includes an extra field to indicate if the block is the last in a sequence. r_wrap doesn't kill the fork until we reach the last block in the sequence.
- Added testcase big-line.sh in evaluation/round_split/scrips to demonstrate the memory usage benefit.


evaluation/round_split/scripts/sort-sort.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
angelhof, line 1: Is it possible that you use the scripts and inputs from benchmarks instead of copying them again under round_split?

angelhof:
Also there seems to be a compiler warning for line 41 of r_split.c that bufLen is unused. If it should not be used it should be deleted.

angelhof (approved):

PR #229, opened 2021-04-20
Make the bigrams test more realistic

angelhof:


nvasilakis (approved):
Little idea: here would be to {merge this with, use cases from} oneliners and keep only non-benchmark tests, because this duplication might cause minor issues (for example, bigrams here is different from the one in benchmarks).

PR #231, opened 2021-04-20
Comment out set -e in poets

angelhof:
Currently PaSh does not handle `set -e` that well and so 6_7.sh in poets when run in bash sequentially exits after the first grep if it finds nothing, while PaSh keeps executing. Commenting out set -e for now for the tests to run and we can comment it back in when we fix the set -e behavior.


nvasilakis (approved):
Ah great. I'll add this somewhere in the documentation.

PR #232, opened 2021-04-20
bash-c fix escape characters in r_wrap

tammam1998:


compiler/ir_utils.py
```diff
@@ -143,6 +143,29 @@ def format_expanded_arg_char(arg_char):
         ## TODO: Make this correct
         raise ValueError
 
+def format_arg_to_raw_chars(arg_char_list):
+    chars = [format_arg_to_raw_char(arg_char) for arg_char in arg_char_list]
+    return "".join(chars)
+
+def format_arg_to_raw_char(arg_char):
```
angelhof, line 150: I don't see the difference of this from this: `format_arg_char`.

compiler/definitions/ir/nodes/r_wrap.py
```diff
@@ -28,8 +28,10 @@ def wrap_node(node):
 
     ## TODO: All arguments must be options, otherwise there must be
     ##       special handling in the wrap node2ast code. 
-    old_options_strs = [str(opt) for i, opt in node.com_options]
-    wrapped_command = " ".join([str(node.com_name)] + old_options_strs)
+    # print("options: ",  node.com_options.replace("\\", ""))
+    
+    old_options_strs = [opt.raw_str() for i, opt in node.com_options]
+    wrapped_command = " ".join([node.com_name.raw_str()] + old_options_strs)
     wrapped_command_arg = [(1, Arg(string_to_argument(f"\'{wrapped_command}\'")))]
```
angelhof, line 35: I would prefer not going back to string to go back to argument. This introduces an unnecessary fail point.

Could we maybe just put all of those in `Q` straight from `Arg`? Argument has an `ast_list` field that contains characters of the form https://github.com/andromeda/pash/blob/main/compiler/parser/ast_atd.atd#L52. 

If this does not work, we need to _at least_ write some comments here about this raw_str and what it does.

compiler/ir_utils.py
angelhof: If there is a difference we need to be explicit and write it in a comment.

tammam1998: This function calls format_arg_to_raw_char which doesn't add escape characters

compiler/definitions/ir/nodes/r_wrap.py
tammam1998: Using Q didn't work because I want the type of quotes to be single ones so that they don't interfere with double quotes that exist in the command itself. I get bash -c "sed"s/T..:..:..//"" instead of 'sed"s/T..:..:..//"'

compiler/ir_utils.py
tammam1998: Rewrote the solution to not more functions

angelhof (approved):

PR #236, opened 2021-04-28
Braces in printing

angelhof:
Python printing causes an issue when running `https://github.com/andromeda/pash/blob/fix-printing-braces/evaluation/intro/demo-spell.sh` because it does not properly add the braces after the `&&` in line 7. Therefore, `exit` in Line 9 is always executed!

@mgree and @thurstond is there any issue with adding braces around everything in the printer? Could we add as many braces to be safe?

### DIFF
--- a/compiler/parser/ceda/ast2shell.py
+++ b/compiler/parser/ceda/ast2shell.py
@@ -177,21 +177,21 @@ def to_string (ast):
 
             return parens (to_string (a) + string_of_redirs (redirs));
         elif (type == "And"):
-            (a1, a2) = params;
+            (a1, a2) = params
 
             return braces(to_string(a1)) + " && " + braces(to_string(a2))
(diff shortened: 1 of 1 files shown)


angelhof:
Note: We might have broken this when we (me and @mgree ) replaced `;` with `\n` in printing.

mgree:
Arbitrary braces should be fine; you need the final brace on its own line (or a `;` first). I'd do it for:

`!` `&&` `||` `;`

There _should_ already be braces logic for other compound commands (e.g., `Bg`/`&`).

angelhof:
Thanks! I will to add to all the suggested spots :)

angelhof:
@nvasilakis if it looks fine squash and merge :)

nvasilakis (approved):

PR #237, opened 2021-04-28
Put mkfifos in temporary dir and make parallel script slightly cleaner

angelhof:


compiler/definitions/ir/file_id.py
```diff
@@ -57,7 +60,8 @@ def to_ast(self):
         ##
         ## TODO: I am not sure about the FileDescriptor resource
         if(isinstance(self.resource, EphemeralResource)):
-            string = "{}#file{}".format(self.prefix, self.ident)
+            suffix = "{}#file{}".format(self.prefix, self.ident)
```
nvasilakis, line 63: Since these are FIFOs, can we name them something like `fifo`, `channel`, or `stream` rather than `file`?

(minor/I can do it)

angelhof: Good point :) I am doing it now!

angelhof: Done!

nvasilakis (approved):
Great!

Issue #238, opened 2021-04-29
Properly use the dash parser to return text even if file does not end with newline


angelhof:
Currently, the dash parser internal states goofs a bit when the input file does not end with a newline.
We address this by copying the input file to a different one and ensuring that it is terminated with a newline.

We should make sure that we handle that in our inteface with dash (https://github.com/andromeda/pash/blob/main/compiler/parser/ceda/parse_to_ast2.py).

@thurstond proposed the following solution in pseudocode:

```
prev_ast = None, prev_linno_before = -1, prev_linno_after = -1;
while (True):
    (cur_ast, cur_linno_before, cur_linno_after) = ...; # The existing ctypes stuff

    if (nleft == EOF_NLEFT): # -99, defined in libdash/src/input.c
        assert (n_ptr_C == NEOF);
        # Last line didn't have a newline but it was parsed so we still need to count it
        prev_linno_after ++;
    else:
        assert (nleft == 0); # In all other cases, dash parses whole lines.

    if (prev_ast != None):
        yield (prev_ast, prev_linno_before, prev_linno_after);

    (prev_ast, prev_linno_before, prev_linno_after) = (cur_ast, cur_linno_before, cur_linno_after);

    if (n_ptr_C == NEOF):
        yield (cur_ast, cur_linno_before, cur_linno_after);
```

thurstond:
I was stupidly reading nleft before parsecmd_safe. When reading it afterwards, the logic to handle EOF is simple:

https://github.com/andromeda/pash/commit/f5197583fe8222f6d48320144a07063a77a0134a

PR #239, opened 2021-04-29
Extend PaSh to not unparse an AST if it is not modified at all, but rather print it out as is

angelhof:
This requires a small change to the way we use the dash parser, asking it for the lines that it has read.

To make that work, we needed to add a step that copies the input file to a temporary file so that it makes sure that it contains a newline before EOF. Since the current version of the dash parser is a bit goofy there.

Eventually, we should make this solution slightly better, by calling the interface properly (some hints in #238).


compiler/parse.py
```diff
@@ -6,23 +6,48 @@
...
+    shell_list = []
+    for ast in asts:
+        if(isinstance(ast, UnparsedScript)):
+            # log("Unparsed script is going back out")
+            # log(ast.text)
+            # exit(1)
+            shell_list.append(ast.text)
+        else:
+            ## We are working with two different abstractions for ASTs, one is the class and the other
+            ## is its JSON object form. Due to Python's _disgusting_ lack of types (and our bad code)
```
mgree, line 35: 😂

compiler/parser/ceda/ArgCharList.c
```diff
@@ -0,0 +1,118 @@
+#include <malloc.h>
```
mgree, line 1: Kind of confused why these are in the diff... are they actually new, or just moved?

compiler/parser/ceda/ArgCharStack2.c
```diff
@@ -0,0 +1,125 @@
+#include <string.h>
```
mgree, line 1: Is there an ArgCharStack1.c or ArgCharStack.c? I'm even more confused...

compiler/parser/ceda/Makefile
```diff
@@ -10,7 +10,7 @@ all: parse_to_json2 json_to_shell2 prettyprint_json libdash2.so
 
 
 # C binaries
-parse_to_json2: parse_to_json2.o ast2a.o ast2json.o dash2.o CharList.o Stack.o arg_char.o
+parse_to_json2: parse_to_json2.o ast2a.o ast2json.o dash2.o Stack2.o arg_char.o ArgCharStack2.o
```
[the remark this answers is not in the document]
angelhof: @thurstond Do you have any idea? I think all of these changes happened in the first commit (which is the optimization of the C implementation)/

compiler/parser/ceda/parse_to_json2.c
```diff
@@ -74,9 +76,11 @@ int main (int argc, char* argv []) {
         }
     }
 
+for (int i = 0; i < 1; i++) {
```
mgree, line 79: ?

compiler/parser/ceda/parse_to_json2.py
```diff
@@ -15,15 +16,26 @@
...
 
-    json = serialize_asts_to_json (new_asts)
+    json = []
+    for output in (parse_to_ast (inputPath, True)):
+        (new_ast, verbatim, linno_before, linno_after) = output;
+        json.append (new_ast);
 
-    print (json)
+        # Debugging
+        if (False):
```
mgree, line 32: Any reason not to change this to `log`?

compiler/pash.py
```diff
@@ -79,30 +77,65 @@ def parse_args():
...
-            ## If the shell is invoked with -c and arguments after it, then these arguments
-            ## need to be assigned to $0, $1, $2, ... and not $1, $2, $3, ...
-            if(len(args.input) > 0):
-                ## Assign $0
-                args.name = args.input[0]
-                args.input = args.input[1:]
-            args.input = [fname] + args.input
-
-    if (len(args.input) == 0):
+            ## If last line does not contain a newline before EOF, we add one since the dash parser is slightly goofy.
```
mgree, line 92: Elaborate? "goofy in how it treats commands that end with EOF instead of EOL".

compiler/util.py
```diff
@@ -9,6 +9,11 @@
 def flatten_list(lst):
     return [item for sublist in lst for item in sublist]
 
+def unzip(lst):
```
mgree, line 12: `list(zip(*lst))`? Sorry for the code golf. (Will be degenerate on empty lists, returning `[]`.)

mgree (commented):
Loooks good to me!

I'm a little confused by some of the file renames, but NBD.

A few other comments in there.

compiler/parser/ceda/Makefile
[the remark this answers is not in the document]
thurstond: These C files are part of the C port, and not needed for the Python port.
(Technicality: an older version of the Python port called the dash2.c shims, but that is no longer the case.)

compiler/util.py
angelhof: I tried that first, but the problem is that it returnes a list of tuples, and not lists. This in turn led to some error somewhere (that I am sure can be fixed but I didn't want to spend the time on it) :)

compiler/parser/ceda/parse_to_json2.c
angelhof: @thurston ?

compiler/parser/ceda/ArgCharStack2.c
angelhof: @thurston ?

compiler/parser/ceda/ArgCharList.c
angelhof: @thurstond ?

thurstond: This is part of the C port. It does not affect the Python port.

It's an old change that created a specialized data structure for ArgCharList instead of overloading CharList (C doesn't have templates).

compiler/parser/ceda/ArgCharStack2.c
thurstond: This is part of the C port. It does not affect the Python port. The change is appearing now because of the vagaries of git branching.

Somewhere in the ceda branch of git, there is ArgCharStack.c, which used a linked list (the "natural" translation of the OCaml list code). This optimized version uses an expanding array instead.

compiler/parser/ceda/parse_to_json2.c
thurstond: This is part of the C port. It does not affect the Python port.

I accidentally let through part of the instrumentation for benchmarking (here it parses once; I also time it parsing 100x, etc.).

compiler/parse.py
angelhof: this code = 💩 

I will open an issue to properly address this when I have some time :)

nvasilakis (approved):

PR #242, opened 2021-04-30
Address incompatibility of shift

angelhof:
Running `./pa.sh --dry_run_compiler -c 'shift; echo $1 $2' pash 2 3 4 5` returns `2 3` compared to running `bash -c 'shift; echo $1 $2' pash 2 3 4 5` that returns `3 4`. The issue is that the effect of `shift` when executed through pash's runtime does not propagate outside.

This PR fixes it by propagating shift changes in and out of the executed scripts.


nvasilakis (approved):
Awesome!

PR #245, opened 2021-05-05
Dgsh buffer size

tammam1998:
- Added dgsh_tee buffer size configuration inside config.yaml
- Changed sigpipe to sigkill because dgsh ignores sigpipe


tammam1998:
should merge #232 first

runtime/dgsh_tee.sh
```diff
@@ -8,7 +8,7 @@ PASH_TOP=${PASH_TOP:-$(git rev-parse --show-toplevel)}
 
 cleanup()
 {
-    kill -SIGPIPE $dgsh_tee_pid > /dev/null 2>&1
+    kill -SIGKILL $dgsh_tee_pid > /dev/null 2>&1
```
angelhof, line 11: Maybe we want `SIGTERM`, which is more standard termination.

nvasilakis: Minor comment: I agree too with going with `SIGTERM`, because it gives some time for the process itself to exit gracefully, whatever that means for each process. As commands be have `SIGTERM` handlers though (possibly ignoring it/blocking it),  we have to be prepared to do another final round of cleanup with `SIGKILL`.

angelhof: I agree with the comment about handlers but note that this `kill` here is only performed on `dgsh-tee` and therefore is not meant to be generally applicable/generalizable.

angelhof (approved):

PR #247, opened 2021-05-05
Fix pash's handling of $0

angelhof:
This is a WIP PR that fixes PaSh's handling of `$0` both when calling it with `-c` and when calling it with a normal file.

#246 needs to be merged first because `$0` need to be output correctly by PaSh's expansion too.


angelhof:
Merging without approval just to be able to work on #246 withoutbeing blocked.

PR #248, opened 2021-05-06
[DRAFT] Parsing assertion failing

angelhof:
The parsing assertion https://github.com/andromeda/pash/blob/main/compiler/parser/ceda/parse_to_ast2.py#L72 fails when executing PaSh on the simple `heredoc1.sh` that is shown below. Note that the script does not end with a newline before EOF.

@thurstond is it reasonable that this assertion fails there?

@mgree what do you think?

Note that this assertion failing is behind 8 POSIX test fails.

### DIFF
--- a/evaluation/tests/interface_tests/heredoc1.sh
+++ b/evaluation/tests/interface_tests/heredoc1.sh
@@ -1,0 +1,3 @@
+cat << foo
+line one
+foo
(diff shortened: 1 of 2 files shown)


angelhof:
FYI, by commenting out this assertion all pash tests `./scripts/run_tests.sh` pass and we pass 5 more POSIX tests.

thurstond:
Somehow, the parser logic did increment the final line number even though the file doesn't have a trailing newline.

I don't know how many special cases there are, or if we can even figure them out easily (without inspecting the AST), so it might be easiest to relax the assertion:

`
assert ((linno_after == len (lines)) or (linno_after == len (lines) + 1));`

angelhof:
I made the change and we can probably keep it like this for now until an issue arises.

PR #249, opened 2021-05-07
Loop argument parsing issue

angelhof:
The `loop1.sh` testcase (simplified from a POSIX test) is parsed wrongly by pash which only parses the first loop argument `A` and not `B`.

One can reproduce this by running `$PASH_TOP/pa.sh -d 1 loop1.sh` with the added logs in this branch to see that it is a parsing issue.

@thurstond, @mgree do you have any ideas what could be the problem?

### DIFF
--- a/compiler/ast_to_ir.py
+++ b/compiler/ast_to_ir.py
@@ -492,10 +492,9 @@ def replace_ast_regions(ast_objects, irFileGen, config):
     preprocessed_asts = []
     candidate_dataflow_region = []
     for i, ast_object in enumerate(ast_objects):
-        # log("Preprocessing AST {}".format(i))
-        # log(ast_object)
+        log("Preprocessing AST {}".format(i))
+        log(ast_object)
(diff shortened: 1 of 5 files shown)


thurstond:
I've just committed 2476ec0c6d78836dd2cd3eb7f5e0bd4cb894ac29, which changes the for loop parsing (and unparsing) from "arg" to "args".

Note that this necessarily breaks the API i.e., for parsed loop1.sh, notice the arguments has an extra layer of nesting:

["For", [1, **[[["C", 65]], [["C", 66]]]**, ["Command", [2, [], [[["C", 101], ["C", 99], ["C", 104], ["C", 111]], [["V", ["Normal", false, "idFor1", []]]]], []]], "idFor1"]]

@mgree If this change is correct, your upstream OCaml code needs a similar tweak.

thurstond:
My branch is slightly out of sync so I will merge the braces change back in.

angelhof:
Thanks everyone!

PR #251, opened 2021-05-11
Escaping unparsing issue

angelhof:


angelhof:
@mgree if everything looks cool for now we can merge :)

angelhof:
@dkarnikis can we figure out what fails in the Ubuntu 18 VM? Could it have to do with the python version in it (and some incompatibility with respect to this subprocess call `https://github.com/binpash/pash/blob/future/compiler/pash.py#L102`).

mgree (approved):
Looks good to me!

angelhof:
@dkarnikis I found it, it had to do with the python version. I fixed it now to work starting from 3.6. Would there be any way to fix the python version in all testing instances to avoid such issues in the future?

dkarnikis:
@angelhof sorry I was afk, I will try to solve this issue tomorrow (:

angelhof:
No worries, I fixed the issue. The thing that we have to figure out is how to fix the python versions that we support and make sure that we only test for them (and not other ones depending on the Ubuntu version) :)

PR #256, opened 2021-06-02
Pip install fixes

dkarnikis:
pip install now install packages on $PASH_TOP/python_deps
The packages needed are: 

- jsonpickle==1.4.2
- PyYAML==5.4.1
- numpy==1.19.5
- matplotlib==3.3.4

After setup-pash.sh is executed, an echo message shows the path the user should export
export PYTHONPATH=$PASH_TOP/python_deps

### DIFF
--- a/pa.sh
+++ b/pa.sh
@@ -2,7 +2,7 @@
 
 export PASH_TOP=${PASH_TOP:-${BASH_SOURCE%/*}}
 export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/"
-
+export PYTHONPATH=${PYTHON_DEPS:-$PASH_TOP/python_deps}
 if [ "$#" -eq 1 ] && [ "$1" = "--init" ]; then
   $PASH_TOP/compiler/superoptimize.sh
   exit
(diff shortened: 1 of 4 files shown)


angelhof:
Thanks! I am not sure if that is the correct way of addressing this issue. Adding something to `PYTHONPATH` is not project local. Is there anyway to really have these dependencies only when running pash? I am not very familiar with "proper" python projects, @abenetopoulos @nvasilakis do you have any idea?

nvasilakis:
@abenetopoulos should know more about this!

dkarnikis:
You were, there is **virtualenv** that creates a specfial environment for 'local' python package installation.
I removed the existing code and switched to virtualenv. Helper messages are shown after setup-pash is executed.
The user has to user the said environment (`source pashenv/bin/activate`) where I have installed the packages.
To leave the pash environment, the user has to execute `deactivate`

angelhof:
Thanks, that looks better! Can any of you two @nvasilakis @abenetopoulos review and confirm that there is nothing wrong with this? :)

nvasilakis:
Looks good. Could we maybe have a `requirements.txt` with contents of the form:

```
jsonpickle==1.4.2
PyYAML==5.4.1
numpy==1.19.5
matplotlib==3.3.4
```

and then just do something like:

```
# after we activate the environment
pip install -r requirements.txt
```

Also, was this tested on fresh VMs with the four major distros we support?

dkarnikis:
The machines that I tested were: Ubuntu 20.04, Fedora 34, Arch-linux rolling, Debian 10.
I have added the python changes to [ae-fixes](https://github.com/andromeda/pash/pull/258).
Closing this PR.

PR #263, opened 2021-06-14
Remove old input files on each execution

dkarnikis:


angelhof:
Thanks a lot! This seems to be needed only for `unix50`, which is not idempotent. I created a new issue to keep track of this. The fix is fine for now but we should consider making all `setup.sh` scripts idempotent, e.g., by only making the files larger if they are 1GB in size).

PR #274, opened 2021-07-10
PaSh installation for Windows WSL

dkarnikis:
Step-by-step guide to setup WSL and install PaSh

### DIFF
--- a/docs/windows.md
+++ b/docs/windows.md
@@ -1,0 +1,34 @@
+# Windows Installation
+To setup and install PaSh on a Windows machine, `Windows Subsystem for Linux (WSL)` needs to be installed.
+
+## What is WSL
+The Windows Subsystem for Linux lets developers run a GNU/Linux environment 
+-- including most command-line tools, utilities, and applications -- 
+directly on Windows, unmodified, without the overhead of a traditional virtual 
+machine or dualboot setup.
+
+## Setup WSL
+
+- Open PowerShell as Administrator and run:
+
+    `dism.exe /online /enable-feature /featurename:Microsoft-Windows-Subsystem-Linux /all /norestart`
+
+- Enable Linux Filesystem for Windows
+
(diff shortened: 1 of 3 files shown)


docs/windows.md
```diff
@@ -0,0 +1,34 @@
...
+
+- Open PowerShell as Administrator and run:
+
+    `dism.exe /online /enable-feature /featurename:Microsoft-Windows-Subsystem-Linux /all /norestart`
+
+- Enable Linux Filesystem for Windows
+
+    `dism.exe /online /enable-feature /featurename:VirtualMachinePlatform /all /norestart`
+
+- Install the distro of your choice:
```
nvasilakis, line 20: Maybe only mention distros we have tried/supported by PaSh for now?

nvasilakis (approved):
Awesome work, thanks so much!

Maybe add a pointer from the main setup READMEs (top-level pash, and in /doc) to this file so that users find it? Maybe at the point where we have all distros?

docs/windows.md
dkarnikis: Fixed!

docs/windows.md
```diff
@@ -0,0 +1,32 @@
+# Windows Installation
+To setup and install PaSh on a Windows machine, `Windows Subsystem for Linux (WSL)` needs to be installed.
+
+## What is WSL
+The Windows Subsystem for Linux lets developers run a GNU/Linux environment 
+-- including most command-line tools, utilities, and applications -- 
+directly on Windows, unmodified, without the overhead of a traditional virtual 
+machine or dualboot setup.
+
+## Setup WSL
```
angelhof, line 10: I am not sure we should include such information in our documentation. The reason is that these instructions might quickly get outdated. Can we instead point to an official WSL installation link, e.g., https://docs.microsoft.com/en-us/windows/wsl/install-win10

PR #275, opened 2021-07-12
Make WSL2 installation instructions more detailed

janekb04:
Add some missing steps in the WSL2 installation process and describe it more throughoutly.

### DIFF
--- a/docs/windows.md
+++ b/docs/windows.md
@@ -19,7 +19,7 @@ and the [WSL Troubleshooting Guide](https://github.com/MicrosoftDocs/WSL/blob/ma
 All commands below should be run in an Administrator Poweshell prompt.
 
 - If Running Windows 10 Preview build 20262 or higher:
-    - List the available Linux Distributions
+    - List the available Linux Distributions. Note that currently only Ubuntu, Debian and Fedora are supported by PaSh.
         ```powershell
         wsl --list --online
         ```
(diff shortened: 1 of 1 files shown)


docs/windows.md
```diff
@@ -1,32 +1,82 @@
...
+            (New-Object System.Net.WebClient).DownloadFile("https://wslstorestorage.blob.core.windows.net/wslblob/wsl_update_x64.msi", "C:/Windows/Temp/wsl2.msi");
+            msiexec /i "C:/Windows/Temp/wsl2.msi";
+            Remove-Item "C:/Windows/Temp/wsl2.msi"
+            ```
+        - Set WSL2 as default WSL version
+            ```powershell
+            wsl --set-default-version 2
+            ```
+    - Install a Linux distribution from the Microsoft Store (cannot be done using command line, 
+      unless using something like Chocolatey, but in that case the whole process is automated):
```
nvasilakis, line 59: Here we can only include distributions for which we have verified that PaSh runs (currently these are Ubuntu 18.04 and 20.04, Fedora, Debian, and Arch.)

nvasilakis (approved):
Thanks so much for this, @janekb04! Feel free to merge (using squash) after you make the changes:-)

janekb04:
@nvasilakis I don't think I have the privileges to merge right now. I also won't need to squash, as the aggregators will be in a separate PR to keep things clean.

nvasilakis:
Ah, thanks for letting me know @janekb04 , I did not know you don't have the privileges. I will take care of merging then:-)

Would it be easy to [address the comment in my review](https://github.com/binpash/pash/pull/275#discussion_r668086895)? I will then merge, thanks so much!

docs/windows.md
janekb04: Ok, I removed the untested distros. Sorry, I didn't notice your comment here, but GitHub only notified me about your top-level comment.

nvasilakis: No worries, and thank you!

angelhof:
@janekb04 @nvasilakis I think we should not copy WSL installation instructions in PaSh's documentation. A link to the official instructions is better (since otherwise we would have to maintain a copy of WSL installation instruction making sure that it doesn't become stale).

nvasilakis:
@angelhof agreed! Documentation is getting a serious overhaul, will fix accordingly!

PR #261, opened 2021-06-09
CI/CD test parsing fixes. WIP

dkarnikis:

PR #277, opened 2021-07-13
Add C++ aggregators

janekb04:
Use C++ aggregators instead of Python aggregator implementations. In the future, there will be additional ones. Currently, the aggregators are `make`d when invoking `scripts/setup-pash.sh`. Tests have been moved from `annotations/tests/` to `runtime/agg/cpp/tests` to be next to the aggregators being tested. They have been set up to be invoked when calling `scripts/run_tests.sh`. Currently, no tests are implemented as 1. no real aggregators were implemented so far (only for `grep` and it seems to be using built-in ones) and 2. I wasn't able to find any code in the PaSh codebase that parses the `aggregate` member in annotations' JSON and could be used to extract information about which aggregator corresponds to which command.
## To-Do List:
Implement aggregators, add them to Makefile, create tests for the implemented aggregators, and change annotation files for the following commands:
- [x] ~~`nl`~~ _This command should not be parallelized, as per [this comment](https://github.com/binpash/pash/pull/277#issuecomment-879783363)_
- [x] `wc` including `-lwcmL`
- [x] `uniq` including `-c`
- [x] `tr` including `-ctd`
- [x] `grep` including `-cEFGPefiwxsvoqaI`


janekb04:
Aggregators yet to be added include (for future reference): `nl`, `wc`, `uniq -c`, `tr -cs`.

nvasilakis:
@janekb04 This is awesome! I will pull, test, and report back later today—but this looks so exciting!

Curious: why delete the testing infrastructure? These scripts were pretty generic and used the annotations to identify the right aggregator path.

janekb04:
@nvasilakis I deleted the test infrastructure as it was rather basic and didn't have support for commands that require different aggregators depending on their options. I think that for now, after I implement some aggregators to test, I will write manual tests and when infrastructure used to parse `aggregate` members will be in place, I will make the tests use it instead.

nvasilakis:
OK:-)

janekb04:
Now that the base infrastructure is in place, I think that what is left is for me to do, is to add missing aggregators, change annotation files to point to those aggregators, add tests for those aggregators.

nvasilakis:
> Now that the base infrastructure is in place, I think that what is left is for me to do, is to add missing aggregators, change annotation files to point to those aggregators, add tests for those aggregators.

Sounds good!

janekb04:
Regarding the failing CI, it seems like the tests are fine and the problem is with the workflow not being able to post the results somewhere.

dkarnikis:
Yeah, the CI is not yet complete, that's why it fails

janekb04:
@nvasilakis  Now, while working on aggregators, I've come to the realization that I don't think that parallelizing certain commands makes sense. 

Let's take `nl` as an example. If not parallelized, it has to go through a file in `O(n)` time and add line numbers. If it is parallelized than it can be spread between `k` threads and take `O(n/k)` time in theory. However the aggregation step has to go through each of the partial results sequentially and merge them in `O(n)` time, so we get the original time complexity. Actually, the added split and `nl` commands will make this approach slower than just using only one invocation of `nl`.

I would conclude from this that parallelizing only makes sense if...
- let `cmd` be a shell command
- let `k` be the number of threads (I believe this is `--width` in PaSh)
- let `P(x)` be the time complexity of `cmd` on an input of size `x`
- let `Q(x)` be the time complexity of `cmd`'s aggregator on an input of size `x`

...if  `P(x)/k + Q(x) < P(x)`. That means that `Q(x) < P(x) - P(x)/k`, which could be approximated to (assuming a large `k`) `Q(x) < P(x)`. The approximation strengthens the fact that it should be `<` rather than `≤`.  From this, I conclude that parallelizing `nl` is counterproductive, as for it `P(x) = x` and `Q(x) = x`, so the inequality is not satisfied.

Is this right, or am I getting something wrong?

nvasilakis:
This is right!

janekb04:
Does PaSh, then, have some check implemented so it doesn't paralelize such commands?

janekb04:
I don't recall there being anything about this in the paper. Perhaps this could be causing the cases where PaSh underperforms, like when it has 0.89x speedup?

nvasilakis:
The base information is captured by annotations (and the logic that implements associated transformations). There is a higher-level, more general question about when to parallelize combinations of commands and this includes several parameters — commands before/after, input sizes, target environment etc. This is not fully implemented yet.

janekb04:
I completed the aggregator for `wc`. I also changed the code a bit to use streams instead of strings and added test infrastructure and changed the annotation file. Tests pass for me.

janekb04:
This new testing infrastructure doesn't ignore whitespace. So the `wc` aggregator finds the padding in the output of `wc`.

janekb04:
@nvasilakis Now the aggregator for `uniq -c` has also been added. It passed the tests on my machine. It assumes that the split files end in a `\n` character, but I think that this assumption is always met when PaSh splits files.

janekb04:
@nvasilakis Now that the `wc` and `uniq -c` aggregators have been implemented, which aggregators should I implement next? I know there is `tr -cs` however on the last technical meeting we discussed how currently the aggregators don't have access to the command's arguments and in this case that is a requirement for the aggregator. Is there a list of aggregators to implement or perhaps a list of most common commands that need aggregators?

janekb04:
Also, it seems like the CI isn't working yet. I'm not sure if #261 was supposed to fix it, but I'm mentioning it just in case.

dkarnikis:
@janekb04 I suppose that I have fixed the CI issues. I am not quite sure whether the tests will be executed on this pr (since your fork/pr is older than my patch). I guess we will find out soon

janekb04:
@dkarnikis I reran the tests and they still seem to fail in the "Posting results" section.

dkarnikis:
Probably because your fork/pr is based on an older version of pash 
(on newer prs, you should not have this issue as long as your fork is updated to match the main repo).

Based on your execution logs, the tests seem to run correctly.
```
intro: 2/2 tests passed.
interface: 6/6 tests passed.
compiler: 52/52 tests passed.
```

dkarnikis:
Ok that's much better now. As mentioned in this #286, a previous commit broke the scripts

janekb04:
@nvasilakis I added support for options for aggregators. Now aggregators like `uniq` will be able to see if an option such as `-c` has been set and modify their behaviour accordingly. I think this will prove especially useful with commands that have a lot of options, such as `sort`. 
It took me some time as the Linux `getopt_long` command isn't very clean to use, so I wanted to make a clean C++ wrapper around that. I also didn't want to sacrifice any performance, so all data structures for `getopt_long` are prepared at compile time using some template metaprogramming.

janekb04:
Fortunately, after all that, thanks to using `getopt_long`, now all syntaxes like `-c -s`, `-cs`, `-Wall`, `-opt -- other`, `--combine` should be correctly parsed.

janekb04:
I added all options (`-lwcmL`) for the `wc` aggregator. I also added VSCode, Clion and Visual Studio files to `.gitignore` because it was annoying for me to constantly delete them.

nvasilakis:
> Fortunately, after all that, thanks to using `getopt_long`, now all syntaxes like `-c -s`, `-cs`, `-Wall`, `-opt -- other`, `--combine` should be correctly parsed.

Awesome!

janekb04:
@nvasilakis Support for `uniq` (with no `-c`) and empty input files has been added. I think this no longer has to be a draft.

janekb04:
@nvasilakis I just wanted to mention here, just in case, that I added a concatenating aggregator for `tr` and its options `-ctd`. I haven't added an aggregator for `tr -s` yet, as I'm unsure about some license issues I wrote about in an email.

janekb04 (commented):
I think the PR is ready. All aggregators compile (when running `setup_pash.sh`), all unit tests succeed (`when running `run_tests.sh`) and the base infrastructure for implementing future aggregators is in place.

janekb04:
Also, I don't think we should squash when merging. The PR is quite large and I think the iterative change process should be preserved.

dkarnikis:
@janekb04 could you fetch the latest changes from the main repo (as you did before) so the tests may run properly? Also could you follow the steps on the DCO check that fails? 
Thanks

janekb04:
@dkarnikis I already fetched. I don't know why the tests fail. I will look into DCO.

janekb04:
@dkarnikis It look like tests fail on "building runtime". This could possibly be caused by aggregators not compiling, but on my machine it works fine. I'll investigate.

janekb04:
It works on Ubuntu 20.04 ([link](https://github.com/binpash/pash/pull/277/checks?check_run_id=3154309751#step:3:28)). I think the issue is that the compiler on 18.04 doesn't support C++17. Can I just add a `sudo apt install g++10` to `setup_pash.sh`?

dkarnikis:
That would be a solution. However, similar changes should be done in order to support the other distros (fedora, arch, debian) but also their docker files.

janekb04:
@dkarnikis Does it have to support any other platform/distro other than Ubuntu, Fedora, Arch, and Debian? Does it have to support macOS?
I never used Docker before. I can begin learning it though, if necessary.

dkarnikis:
Currently these are the only distros we support. I am working on Mac 'docker' support. Maybe you could create an issue to fix the Dockerfiles to support the new aggregators. @nvasilakis  what do you think?

janekb04:
@dkarnikis I run the instructions on DCO.

nvasilakis:
We could also update the dockerfiles after Jan pushes changes to the setup
scripts.

janekb04:
I will now push a test commit to see if DCO and signing works.

janekb04:
I realized that I signed the previous commits with the wrong email for DCO. I will resign them now and run the DCO commands again.

janekb04:
After doing the DCO stuff I realized that some files have changed that I haven't touched myself. I think they come from a different PR. Here's the list of the files I personally didn't change:
```
.github/workflows/cicd.yaml
.github/workflows/main.yml
README.md
annotations/README.md
annotations/cat.json
annotations/diff.json
annotations/echo.json
annotations/find.json
annotations/head.json
annotations/history.json
annotations/pr.json
annotations/pwd.json
annotations/tac.json
annotations/tail.json
compiler/test_evaluation_scripts.sh
docs/README.md
docs/contributing/contrib.md
docs/pash_architecture.PNG
docs/tooling/ci.md
docs/tutorial/tutorial.md
docs/windows.md
evaluation/benchmarks/bio/README
evaluation/benchmarks/max-temp/max-temp.sh
evaluation/benchmarks/poets/8.2_2.sh
evaluation/benchmarks/run.par.sh
evaluation/benchmarks/run.seq.sh
evaluation/benchmarks/run_all_benchmarks_ci.sh
evaluation/intro/test.sh
evaluation/tests/input/setup.sh
evaluation/tests/interface_tests/run.sh
scripts/distro-deps.sh
scripts/docker/Dockerfile.base
scripts/docker/Dockerfile.testing
scripts/docker/README.md
scripts/docker/debian/Dockerfile
scripts/docker/fedora/Dockerfile
scripts/docker/ubuntu/Dockerfile
```

janekb04:
I updated `distro_deps.sh`. By the way, it took me so long because I couldn't figure out why `git` wasn't able to sign my commit. Apparently the issue turned out to be that I was writing `git commit -Ss -m "message"`, while it should've been `git commit -sS -m "message"`...

janekb04:
@nvasilakis @dkarnikis The PR is ready for review. I changed `distro_dependecies.sh`, so now the aggregators compile. I also reverted annotations as discussed with Nikos.
I don't know why the CI fails. But it doesn't seem to be related to aggregators. The aggregators build successfully and all unit tests pass.
I didn't delete the `tr` aggregator, as it will make it easier to implement `tr -s` in the future.

janekb04:
@nvasilakis @dkarnikis I will also write here, just in case, that we discussed with Nikos that the PR won't be squashed as it affects too many files.

janekb04:
@nvasilakis What’s the current state of is PR? Should I wait for it to be reviewed or should I continue implementing more aggregators and push them?
I don’t want to open a new PR before this one is merged because it would need the base infrastructure from this PR anyway.

janekb04:
I have added the aggregator for `grep`, including many of its multiple options.

When looking into the source code you may find that it mostly consists of just option names. This is on purpose.

The aggregation itself is quite simple, just concatenation or addition of two numbers.

Despite that, the aggregator infrastructure still requires to specify all supported options regardless of the fact wether they change the aggregation process logic. I decided to do it this way to follow PaSh's philosophy of "correctness first". I want it to be needed to explicitly specify all supported options in order not to run into a situation when the aggregator would fail. It also encourages to think about every option and whether it is actually supported.

janekb04:
@nvasilakis After looking through the files again, I don't think there's anything broken apart from `head.json`. Now, all changes left are those made intentionally by me. As such, I believe we can merge this branch without recloning. I don't really know the `git` tools related to squashing, reapplying or modifying commits in any way.

I will recap for each file, why it changed:

 - `.gitignore` ignores files of IDEs like VSCode, Clion or Visual Studio - makes it less annoying while developing
 -  **_`annotations/head.json` this is the only one, I didn't change_**
 - `annotations/tests/*` directory deleted as tests are now in `runtime/agg/cpp/tests`
 - `compiler/parser/Makefile` I misremembered. I did change this file. It shouldn't affect anything. I just made it use `$(MAKE)` instead of `make` because I believe this is good practice.
 - `runtime/Makefile` now compiles the C++ aggregators
 - `runtime/agg/README.md` now refers to new C++ aggregators
 - `runtime/agg/cpp/*` these are all new files
 - `scripts/distro-deps.sh` now installs `g++-10` as it is required for compiling the aggregators. Note: I tested it only on Ubuntu. I don't know if it works on Debian, Fedora or Arch. I only know that on Ubuntu 18.04 it wouldn't compile if I didn't change this file.
 - `scripts/run_tests.sh` now runs the aggregator unit tests. Note: they don't affect the CI status, as I didn't really look into how to do that and I'm not sure if it would be beneficial/necessary.

janekb04:
@nvasilakis I fetched upstream and fixed `head.json`. I think the branch is ready for merge.

janekb04:
@nvasilakis I think something is wrong with the CI as it has been continuously running for over two hours now.

dkarnikis:
@janekb04 I cancelled the CI and I am checking the logs. 
Apparently, the setup scripts fails to download the required files (this also happened locally to my setup).
`curl: (7) Failed to connect to ndr.md port 80: Connection timed out`.

@nvasilakis I think you should check this.

janekb04:
@nvasilakis What is the current state of this PR?

compiler/parser/Makefile
```diff
@@ -41,7 +41,7 @@ libdash:
 	cd libdash && ./autogen.sh && ./configure && make
 
 libdash-ocaml:
-	cd libdash/ocaml && make && make install
+	cd libdash/ocaml && $(MAKE) && $(MAKE) install
```
nvasilakis, line 44: Separate commit

scripts/distro-deps.sh
```diff
@@ -28,27 +28,29 @@ case "$distro" in
...
    arch*) 
-    echo "Updating mirrors"
-    pacman -Sy &> $LOG_DIR/pacman_update.log
+     echo "Updating mirrors"
+     pacman -Sy &> $LOG_DIR/pacman_update.log
      echo "|-- running pacman install...."
-    yes | pacman -S git libtool m4 automake curl pkg-config python-pip libffi make autoconf gcc sudo inetutils bc
-    ;;
+     yes | pacman -S git libtool m4 automake curl pkg-config python-pip libffi make autoconf gcc10 sudo inetutils bc
+     ;;
```
nvasilakis, line 54: Separate commit.

nvasilakis (commented):
Looks great

janekb04:
Closed via #310

PR #280, opened 2021-07-14
Annotation updates

DanielSongShen:


annotations/pr.json
```diff
@@ -0,0 +1,36 @@
...
+            	"operands":
+            	[
+            	    {
+            		"operator": "exists",
+            		"operands": ["-T, --omit-pagination"],
+            	    },
+            	    {
+            		"operator": "!exists",
+            		"operands": ["integer"],
+            		"comment": "Unsure how to express this kind of thing. Want to say: operand does not contain an integer."
```
nvasilakis, line 18: Can you add the word `TODO(@DanielSongShen)` here? And also add this filename and line into the README.md file, in a section called `## Issues` or something? Thanks!

annotations/cat.json
```diff
@@ -1,5 +1,6 @@
 {
     "command": "cat",
+    "aggregate": "$PASH_TOP/runtime/agg/py/cat.py",
```
nvasilakis, line 3: This raises an interesting question. Stateless commands generally take `cat` as an aggregator (and thus we don't need to take note of the `aggregate` record—we know it's going to be something like `cat.py` (and PaSh internally emits the `cat` command which has significantly better performance than `cat.py` ). However, a stateless command might have flags that change the command behavior to parallelizable pure — for example the `cat -n` command is pure. As a result, the annotation for the `cat` command needs an `aggregate` record that points to a special aggregator that can deal with the `-n` flag, say `cat-n-aggregator.py`. So it might be that we cannot have a single aggregator per command, but rather an aggregator per set of predicates. 

(This is also related to something @Dixits1 asked/suggested a few meetings ago, on whether we should have one aggregator per command or one aggregator annotation per case.)

nvasilakis (changes requested):
This is great @DanielSongShen ! Could you make please make the changes and push? Thanks!

annotations/cat.json
angelhof: I think I agree that we need an aggregator per predicate and not per command, since the command aggregator might change depending on the flags being used.

nvasilakis: @angelhof Yeah, but remember that aggregators themselves support multiple flags (e.g., the `sort -m` aggregator supports `-r`, `-n`, etc.) and thus their behavior can be configured accordingly. So per command is indeed possible, and might indeed be the cleanest. 

The question, as I see it, is what's more convenient when aggregators overlap. What would I prefer as an annotation developer—to point the command I'm annotating to an existing aggregator (if an existing aggregator works) or to point my predicate to an existing aggregator. I think having one aggregator per command is the cleanest, but introducing overlaps (aliasing) makes things a bit tricky.

nvasilakis: (BTW this discussion probably belongs in an issue somewhere.)

nvasilakis (approved):
Great, thanks!

PR #284, opened 2021-07-16
Update docs

nvasilakis:
Fixes were motivated by the move to binpa.sh.


angelhof (approved):
lgtm. I am still not sure I understand why we need the wsl instructions in our documentation but I don't have a strong opinion. :)

PR #285, opened 2021-07-19
Changed class of shuf command

khushikukadia:


nvasilakis:
Thanks for this PR. `shuf` can be either of stateless or pure, but (i) stateless allows transformations to propagate further, and (ii) pure commands likely need a custom aggregator, which someone would need to write:-) So it might make more sense to keep `shuf` as stateless:-(

PR #290, opened 2021-07-23
Aggregators in annotations

angelhof:
This PR extends PaSh to learn about a command's aggregator from its annotation (previously it was hardcoded).

It also implements it for `sort`.

It makes the following default behavior assumption.

An aggregator inherits all options of the invoked command, possibly adding some new ones.

@nvasilakis is this is a valid default? If we want to change this, the files that are modified in this commit are the only ones that need to be modified.


nvasilakis (approved):
Awesome!<3

PR #293, opened 2021-07-25
bigrams that appear twice Fix

dkarnikis:
First calculate all the bigrams ([4_3.sh](https://github.com/binpash/pash/blob/main/evaluation/benchmarks/poets/4_3.sh) script) and based on the generated output calculate all the bigrams that appear twice


nvasilakis (approved):
Awesome, thanks!

PR #292, opened 2021-07-23
Head annotation fix

angelhof:
Fix the `head` annotation so that tests pass.

PR #295, opened 2021-07-25
Fix correctness failure

DanielSongShen:


DanielSongShen:
Already fixed (https://github.com/binpash/pash/pull/292)

PR #296, opened 2021-07-26
Ignore paths for CI

dkarnikis:
We could/should add more paths/files that should not trigger the CI.
All *.md file changes (even from compiler,runtime,evaluation,annotations) are ignored.
Changes to the benchmarks folder do not trigger the tests.


angelhof:
@dkarnikis do we need this? If not let's delete it.

PR #297, opened 2021-07-26
Add tests for custom aggregators

nvasilakis:

PR #298, opened 2021-07-26
Parallelizable Pure in annotations

angelhof:
This PR modifies PaSh so that all the necessary components of parallelizable pure commands are extracted from annotations (and not being _hardcoded_). The annotation for `sort` is an example of a simple annotation, while the annotation for `bigrams_aux` is one that is complex (contains custom mapper that has more than one output). 

For the simple case where there is no need for a custom mapper, the only requirement is to specify the `aggregator` component of the annotation and make the annotation `parallelizable_pure`.


angelhof:
Merge this after merging #297 since it depends on it.

@nvasilakis feel free to review and merge.

angelhof:
@dkarnikis It seems that the CI was cancelled (it now passes locally so could you check why was it cancelled)? Also could we have a way to run it again without having to do a new commit? No rush, this is not a pressing issue.

dkarnikis:
@angelhof When one of the test fails, the CI cancels the other pending jobs (eg Ubuntu-18.04 was cancelled). In this case 
`ann-agg` failed. By checking the execution log [here](https://github.com/binpash/pash/pull/298/checks?check_run_id=3163080504), it says

```
Executing test: ann-agg
|-- Does not have env file
|-- Executing the script with bash...   exited with 0
|-- Executing with pash --width 2 ...
diff: standard output: Broken pipe
|-- Executing with pash --width 8 ...
diff: standard output: Broken pipe
```

We may re-run the job manually from the same link by clicking `Re-run Jobs` (upper right). [Check here](https://docs.github.com/en/actions/managing-workflow-runs/re-running-a-workflow)

angelhof:
Thanks a lot @dkarnikis. I will try to debug what doesn't work in the ubuntu 20 :)

compiler/command_categories.py
```diff
@@ -104,6 +106,31 @@ def __repr__(self):
...
+## The class that contains the mapper information that was mapped from annotation (rare).
+class Mapper:
+    def __init__(self, mapper_json):
+        ## Exactly one of the two should exist in the JSON
+        assert('name' in mapper_json or 'rel_path' in mapper_json)
+        assert(not ('name' in mapper_json and 'rel_path' in mapper_json))
+        
+        if('rel_path' in mapper_json):
+            ## Set the name to be the absolute path
+            self.name = os.path.join(config.PASH_TOP, mapper_json['rel_path'])
```
nvasilakis, line 118: This is great — it looks like someone can use an absolute path (for testing, if needed), and the resolution would still work correctly! e.g., `os.path.join("/home/nikos", "/Users/nikos/pizza")` would return `'/Users/nikos/pizza'`. So maybe instead of `rel_path`, we could just name it `path`.

nvasilakis (approved):
Looks great!

compiler/command_categories.py
nvasilakis: @angelhof

PR #300, opened 2021-07-28
Add little custom aggregator tutorial

nvasilakis:


annotations/README.md
```diff
@@ -138,7 +138,7 @@ The list identified by "short-long" contains a correspondence of short and long
 
 ## How to Annotate a Command
 
-The first step to annotating a command is to identify its default class: `stateless`, `pure`, `non-parallelizable`, and `side-effectful`. How does the command behave without any inputs?
+The first step to annotating a command is to identify its default class: `stateless`, `parallelizable_pure`, `non-parallelizable`, and `side-effectful`. How does the command behave without any inputs?
```
angelhof, line 141: As you mentioned, we should change `non-parallelizable` here to `pure` (or the opposite).

annotations/README.md
```diff
@@ -178,6 +178,176 @@ For more details, here is an early version of the annotation language:
...
+    [
+        {
+            "predicate": "default",
+            "class": "parallelizable_pure",
+            "inputs": ["stdin"],
+            "outputs": ["stdout"],
+            "aggregator":
+            {
+                "rel_path": "runtime/agg/opt/concat.sh",
+                "options": ["-a"]
```
angelhof, line 305: Is the -a really necessary?

annotations/README.md
```diff
@@ -178,6 +178,176 @@ For more details, here is an early version of the annotation language:
...
+}
+```
+
+**More complex aggregators**:
+Here is the `uniq` annotation that has two cases to support two different sets of flags: 
+
+```
+{
+    "command": "uniq",
+    "aggregate": "$PASH_TOP/runtime/agg/py/uniq.py",
```
angelhof, line 318: This is not necessary.

annotations/README.md
```diff
@@ -178,6 +178,176 @@ For more details, here is an early version of the annotation language:
...
+        }
+    ]
+}
+```
+
+**More complex aggregators**:
+Here is the `uniq` annotation that has two cases to support two different sets of flags: 
+
+```
+{
```
angelhof, line 316: I would again prefer pointing to the real annotations from the tutorial, so that we don't have the risk of lag between the README and the real annotations.

nvasilakis: Ah great idea!

dkarnikis (approved):
Seems fine to me!

PR #304, opened 2021-07-29
Fix space in wf's `tr`

nvasilakis:
Word frequencies seems to have a bug in its invocation of `tr -cs` (a missing space).


angelhof:
Good catch! If tests pass you can merge :)

angelhof (approved):

PR #305, opened 2021-07-29
Fix cut annotation

nvasilakis:


angelhof:
I think that we should not merge this before we investigate what are the new failing test cases. I might have some time to take a look. If we want to merge these (and other changes) faster (without making sure that correctness tests pass) we can use the `future` branch.

nvasilakis:
@angelhof Agreed! BTW we think we know why tests are failing, and we're working to fix them.

annotations/cut.json
```diff
@@ -1,33 +1,56 @@
...
+      "outputs": [
+        "stdout"
+      ]
+    }
+  ],
+  "options": [
+    "stdin-hyphen",
+    "empty-args-stdin"
+  ],
+  "short-long": [
```
angelhof, line 46: This is not yet supported by PaSh. We can open an issue or add a test with a long version of one of these flags and use that in a PR to implement that.

annotations/cut.json
```diff
@@ -1,33 +1,56 @@
...
+        "args[:]"
+      ],
+      "outputs": [
+        "stdout"
+      ]
+    }
+  ],
+  "options": [
+    "stdin-hyphen",
+    "empty-args-stdin"
```
angelhof, line 44: Similarly, this is not implemented at the moment (and is very crucial for all the commands that read from stdin if they have no arguments). In its current form, the annotation leads to test failures because uses of `cut` that read from `stdin` are not supported (that is why previously `cut`'s annotation only reads from stdin).

angelhof (changes requested):

annotations/cut.json
angelhof: This is probably the reason why tests fail.

nvasilakis: OK there are probably multiple reasons why the tests are failing:P

PR #309, opened 2021-08-05
Pash support for FreeBSD

dkarnikis:
Compiler Tests: 46/54 Passed
Interface Tests: 7/7 Passed
Intro Tests :1/2 Passed (due to ndr.pash being down)


compiler/pash_ptempfile_name.sh
```diff
@@ -1,3 +1,22 @@
...
+fi
 
-echo "$(mktemp --tmpdir="$PASH_TMP_PREFIX" -u pash_XXXXXXXXXX)"
\ No newline at end of file
+# convert to lowercase
+distro=$(printf '%s\n' "$distro" | LC_ALL=C tr '[:upper:]' '[:lower:]')
+# now do different things depending on distro
+case "$distro" in
+    freebsd*)  
+        # bsd is so fun :)
```
angelhof, line 13: :')

compiler/pash_ptempfile_name.sh
```diff
@@ -1,3 +1,22 @@
...
+        mkdir -p "$PASH_TMP_PREFIX"
+        tmp=$(mktemp pash_XXXXXXXXXX)
```
angelhof, line 15: Can't we do the same as with the `auto-split` here?

angelhof (approved):
As long as the tests pass this looks good to me :)

compiler/pash_ptempfile_name.sh
dkarnikis: bigrams and alt-bigrams hang on their execution, but I am pretty positive that this is a bash related issue rather than pash. We are going to investigate it with @nvasilakis . All the other tests pass!

compiler/pash_ptempfile_name.sh
dkarnikis: You are correct, I will fix it!

compiler/pash_ptempfile_name.sh
angelhof: great, thanks :)

PR #311, opened 2021-08-06
Port aggregators to FreeBSD

janekb04:
This PR is about porting aggregators to FreeBSD. Todo list (based on #277):
- [x] `wc` including `-lwcmL`
- [x] `uniq` including `-c` *required no changes*
- [x] `tr` including `-ctd`
- [x] `grep` including `-cEFGPefiwxsvoqaI` *required no changes*

Adding support for `sort` will come in a new PR.


janekb04:
More PRs will be created from this branch later, so it shouldn't be deleted after merging.

janekb04:
There is a difference between the Linux version of the `wc` command and the FreeBSD version. The Linux version treats the options `-m`  and `-c` separately and independently. On the FreeBSD version, however they are exclusive and [one cancels the other](https://www.freebsd.org/cgi/man.cgi?wc).

nvasilakis:
Yes. Generally there are quite a few differences between BSD and Linux/GNU Coreutils. A easy way is to use `#ifdef`-style  conditional compilation (i.e. include one kind of functionality if in environment A, a different one if compiled in environment B.) Let me know if this makes sense.

janekb04:
@nvasilakis I am currently finishing the conditional compilation. I am using a different approach though. Instead of using preprocessor directives inside the C++ files I will make the `Makefile` choose a different implementation based on the target system. There will be a separate `.h` file for an aggregator implementation for Linux and a separate file for BSD. If the implementations are the same, then the BSD version will simply consist of one line: `#include ../../linux/cmd/agg.h`.

janekb04:
This one line will suffice. Even no include guards are necessary as the included file already has those and there is nothing else in the file to protect.

nvasilakis:
That's a great idea, @janekb04 !

janekb04:
@nvasilakis I was fighting for a while with the aggregators not wanting to compile on FreeBSD. It turned out that instead of using `make`, `gmake` has to be used. Does PaSh currently automatically use `gmake` on FreeBSD, or should I change PaSh to do so? I am asking because I know that other parts of PaSh also use Makefiles, so this problem would concern not only aggregators, and maybe should be in a different PR?

janekb04:
At this point the `wc` agrregator mostly works on FreeBSD with the exception of some issues with padding. I will analyze [the `wc` source code for FreeBSD](https://github.com/freebsd/freebsd-src/blob/main/usr.bin/wc/wc.c) to investigate.

janekb04:
`wc`, `uniq` and `grep` now work. `tr` is yet to be implemented.

janekb04:
@nvasilakis All aggregators work on Linux and FreeBSD. PR ready for review.

nvasilakis:
Awesome, @janekb04 — Will review and get back to you!

janekb04:
@nvasilakis Thanks. I would also want to remind you, just in case, that I am leaving for the algorithmics "camp" next Monday and then writing the SAT, so I will be away until the end of August since then.

angelhof (approved):
This looks good. Something that we need to think about for the future is that we probably need to figure out a way to disambiguate between the bsd and linux commands in the annotation level too (and therefore by being different commands, their aggregators would be different).

janekb04:
@angelhof The aggregators are always in the `runtime/agg/cpp/bin` directory. Their path doesn't change regardless of the host system, so I don't think that annotation files have to be platform specific.

angelhof:
> @angelhof The aggregators are always in the `runtime/agg/cpp/bin` directory. Their path doesn't change regardless of the host system, so I don't think that annotation files have to be platform specific.

I agree that this is not an issue for aggregators. However, the flags do change behavior and that means that we need to have different annotations per platform. :)

janekb04:
I have verified that aggregators can now compile on Clang including Apple Clang. After some non-extensive initial testing, it seems that in most cases the BSD aggregators seem to be working on macOS out of the box. There are some flags that didn't work yet, but I think that should be addressed by a new PR, so this one can get merged.

PR #319, opened 2021-09-10
PaSh support for interactive execution

angelhof:


angelhof:
The current parser cannot [return lines](https://github.com/binpash/pash/blob/interactive/compiler/parser/ceda/parse_to_ast2.py#L91) (i.e., avoiding unparsing when the whole line cannot be handled by PaSh) when reading from stdin, and therefore interactive mode cannot work at the moment. The parser does return line numbers though. @mgree @thurstond do you have any idea what is the best way to address this? 

A bad solution would be to unparse when doing interactive mode but that is (1) inefficient, and (2) uncovers unparsing issues.

The best solution IMO would be to capture the lines in libdash and return them. @thurstond would that be possible?

Another (more hacky) solution would be to somehow duplicate whatever is read from stdin to memory and then pass it to libdash, making it possible to find the lines using the line numbers returned from the parser.

angelhof:
I completed the interactive infrastructure except for the support for reading from stdin.

angelhof:
I suggest we merge this and then deal with the stdin issue in a separate PR.

thurstond:
> The current parser cannot [return lines](https://github.com/binpash/pash/blob/interactive/compiler/parser/ceda/parse_to_ast2.py#L91) (i.e., avoiding unparsing when the whole line cannot be handled by PaSh) when reading from stdin, and therefore interactive mode cannot work at the moment. The parser does return line numbers though. @mgree @thurstond do you have any idea what is the best way to address this?
> 
> A bad solution would be to unparse when doing interactive mode but that is (1) inefficient, and (2) uncovers unparsing issues.
> 
> The best solution IMO would be to capture the lines in libdash and return them. @thurstond would that be possible?
> 
> Another (more hacky) solution would be to somehow duplicate whatever is read from stdin to memory and then pass it to libdash, making it possible to find the lines using the line numbers returned from the parser.

I'm not convinced there is an easy solution for stdin without duplicating the contents. Consider this example: someone types in one line of 2 billion characters, and dash discovers something it can't parse near the end of that line. We now want all 2 billion characters returned, most of which have already been consumed by libdash already.

angelhof:
> I'm not convinced there is an easy solution for stdin without duplicating the contents. Consider this example: someone types in one line of 2 billion characters, and dash discovers something it can't parse near the end of that line. We now want all 2 billion characters returned, most of which have already been consumed by libdash already.

I think that duplicating contents is an OK first solution. Huge input scenarios could lead to issues but that is pretty rare so we should ignore it for now IMO. Do you think it would be easier to do content duplication on the libdash side or should we do it in PaSh somehow?

mgree:
I think it will be much easier to use `tee` or something than to try to hack libdash (which has a custom lookahead/UNGETC mechanism).

angelhof:
Merging to unblock. Interactive execution works with the exception of the issue with `stdin`, which should be fixed in a different PR.

PR #321, opened 2021-09-15
Fix stdin support for interactivity

angelhof:


angelhof:
There is a potential performance issue with the current hacky solution. It currently `tee`s stdin to both pash and a file (https://github.com/binpash/pash/blob/stdin-interactive/pa.sh#L20) so that the lines of the input can be retrieved if preprocessing determines that pash should not compile that script fragment. However, since this is done unconditionally, it will be done also for scripts where the input is simply data leading to performance overhead.

Does anyone have any idea how to do this `tee`ing simply in python (here: https://github.com/binpash/pash/blob/stdin-interactive/compiler/pash.py#L87) so that we don't have to duplicate the argument parsing logic in `pa.sh` to determine whether we are in interactive mode and only then do the `tee`ing?

angelhof:
I addressed the issue by simply unparsing the parsed script when reading from stdin, since the `keep-unparsed-source` is simply an optimization anyway. Now in cases where we read from a file we can perform this optimization, otherwise not (and we simply unparse).

PR #323, opened 2021-09-16
Add support for break

angelhof:
The issue is that `set -e` leads to pash prematurely exiting when setting the exit code. To avoid this, we perform the exit code set in a checked position (in the condition of an if loop).

Wrong initial explanation follows:

There is an issue with `break` which I think has to do with the fact that when executing it in PaSh, it is not in the context of the `until` loop (rather in the loop, a file is sourced that then breaks).

I am not sure how to address this:
- Either we can somehow reimplement break to break through sourced files
- Or we can avoid preprocessing break (this can not be complete since it wouldn't catch this `var=break; $var`).


angelhof:
merging to unblock

PR #325, opened 2021-09-20
Fix some more unparsing escaping issues

angelhof:


compiler/parser/ceda/ast2shell.py
```diff
@@ -333,7 +333,7 @@ def string_of_arg_char (c, is_quoted=False):
         ## Chars to escape unconditionally
         chars_to_escape = ["'", '"', '`', '(', ')', '{', '}', '$', '!', '&', '|', ';']
```
mgree, line 334: I suspect it would be safe move everything but `"` into `chars_when_no_quotes`, but no sense changing anything if it works.

mgree (approved):
lgtm 😁

compiler/parser/ceda/ast2shell.py
angelhof: Great, I left a comment about that!

PR #330, opened 2021-09-22
Set-e does not apply when executing in the context of `&&` or `||`

angelhof:
PaSh currently sources parts of the script in the context of `&&` and `||` and based on the bash documentation, this prevents `set -e` from working (focus in bold):

> -e
>
>    Exit immediately if a pipeline (see Pipelines), which may consist of a single simple command (see Simple Commands), a list (see Lists), or a compound command (see Compound Commands) returns a non-zero status. The shell does not exit if the command that fails is part of the command list immediately following a while or until keyword, part of the test in an if statement, **part of any command executed in a && or || list** except the command following the final && or ||, any command in a pipeline but the last, or if the command’s return status is being inverted with !. If a compound command other than a subshell returns a non-zero status because a command failed while -e was being ignored, the shell does not exit. A trap on ERR, if set, is executed before the shell exits. 

It is not clear how to fix this since we cannot just `source` because this would lead to a premature exit without restoring state properly. Is that an actual issue though?

### DIFF
--- a/compiler/definitions/ast_node.py
+++ b/compiler/definitions/ast_node.py
@@ -196,5 +196,9 @@ def ast_node_to_untyped_deep(node):
         return [json_key, untyped_json_val]
     elif(isinstance(node, list)):
         return [ast_node_to_untyped_deep(obj) for obj in node]
+    elif(isinstance(node, tuple)):
+        return [ast_node_to_untyped_deep(obj) for obj in node]
+    elif(isinstance(node, dict)):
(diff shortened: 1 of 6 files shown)


angelhof:
So the solution here seems to solve the issue even though it is a bit _scary_. At the moment if `set -e` is set and is executed in a checked position, then PaSh simply exits without finalizing its state (time measurements etc). Surprisingly this does not lead to any issues in the tests and so I think we should go with it.

@mgree what do you think?

mgree (approved):
yolo

PR #331, opened 2021-09-23
Avoid using `-c` when invoking bash in `--batch` mode

angelhof:


compiler/pash.py
```diff
@@ -150,8 +145,8 @@ def parse_args():
     parser.add_argument("--output_preprocessed",
                         help=" output the preprocessed script",
                         action="store_true")
-    parser.add_argument("--interactive",
-                        help="Executes the script using an interactive internal shell session (experimental)",
+    parser.add_argument("--batch",
+                        help="Executes the script using an single shell session instead of parsing line by line and sending to a spawned shell. This was the old default behavior.",
```
mgree, line 149: an single -> a single

angelhof:
Unfortunately, there is an issue. When passing input to the bash process from its stdin, there is the concern that if bash tries to redirect its stdin, e.g., using `exec`, then it will stop receiving input from the original source... 

This was not a problem when executing with `-c` because the source was in a file, so redirecting the stdin didn't matter.

It is not clear to me what is the way forward. 
- If we use `-c` then the exit status in some cases is not equal to the one specified in POSIX tests. 
- If we try to pass the source file (not always possible) as a file then I haven't figured out a way to set the bash name correctly (because it actually executes a different file with the preprocessed script).
- If we pass the input through stdin, then we run the risk of not being able to redirect bash's stdin (since `pash.py` is not tightly integrated with bash but rather interfaces with it through stdin).

My personal preference would be to pick one of the first two (and accept that there will be slight inconsistency with bash, e.g., exit codes might be slightly different in some cases) and then only when necessary use the third option (and in those cases accept that there will be slight issues with stdin redirection).

compiler/pash.py
```diff
@@ -200,6 +195,10 @@ def parse_args():
     elif (len(args.input) > 0):
         shell_name = args.input[0]
 
+    ## Currently pash does not support --batch execution when it receives input from stdin.
+    if (args.batch and len(args.input) == 0):
+        log("Error: PaSh does not yet support --batch and input from stdin!", level=0)
+        exit(1)
```
mgree, line 201: Show a usage message?

compiler/pash.py
```diff
@@ -222,25 +221,21 @@ def shell_env(shell_name: str):
...
         a_flag = '-a'
-    return "exec -a{} bash {} -s $@\n".format(shell_name, a_flag)
+    return "exec -a {} bash {} -s ".format(shell_name, a_flag)
 
 def execute_script(compiled_script_filename, command, arguments, shell_name):
     new_env = shell_env(shell_name)
     subprocess_args = bash_prefix_args()
-    subprocess_args += ["-c", 'source {}'.format(compiled_script_filename), shell_name] + arguments
-    # subprocess_args = ["/usr/bin/env", "bash", compiled_script_filename] + arguments
+    subprocess_args += ["-c", "{} {} < {}".format(bash_exec_string_no_args(shell_name), " ".join(arguments), compiled_script_filename)]
```
mgree, line 238: Do you know about format strings in Python?

You could write:
```
subprocess_args += ["-c", 
f'{bash_exec_string_no_args(shell_name)} {" ".join(arguments)} <{compiled_script_filename}']
```

Not 100% sure that's clearer in this case, but I quite like it, especially when it's just variables being interpolated.

mgree (approved):
Looks good!

mgree:
Just saw your comments. Huh. Right. I mean, in general, line-based mode isn't going to work correctly if bash ever works with input other than commands---so it might not be the right default.

One option for flagless invocation of `bash`---which seems to produce the right answer---is to generate a temp directory, in which we can name the script the exact thing we want. You could then run `(cd $TMPDIR; bash $FILE $ARGS)` or something similar. Awkward, and I'm not sure what else this messes with in the runtime. Instead of running `cd`, you could try locally altering the `PATH`... but both of these will be visible, since cwd is inherited and `PATH` is exported.

Finally, another option is to ask Chet (or search around). I'm surprised by this behavior in `bash`, and it's not obvious that it's intentional.

mgree:
Okay, so `bash` is just super weird. Every other shell behaves the same regardless of where the input is coming from.

```
                smoosh: -c   1 stdin   1 file   1
                  bash: -c 127 stdin   1 file   1
         bash -o posix: -c 127 stdin   1 file   1
                  dash: -c   2 stdin   2 file   2
                   zsh: -c   1 stdin   1 file   1
      zsh --emulate sh: -c   1 stdin   1 file   1
                   osh: -c  33 stdin  33 file  33
                  mksh: -c   1 stdin   1 file   1
                   ksh: -c   1 stdin   1 file   1
                  yash: -c   2 stdin   2 file   2
         yash -o posix: -c   2 stdin   2 file   2
```

Here's the tester script.
```shell
trap 'cleanup' EXIT

cleanup() {
    if [ "$tmp" ] && [ -f "$tmp" ]
    then
        rm $tmp
    fi
}

unset foobar
tmp=$(mktemp)
echo 'echo ${foobar?}' >$tmp

# calculate field width
for SH in smoosh bash "bash -o posix" dash zsh "zsh --emulate sh" osh mksh ksh yash "yash -o posix"
do
    if [ "${WIDTH:-0}" -lt "${#SH}" ]
    then
        WIDTH="${#SH}"
    fi
done
WIDTH=$((WIDTH+6))


for SH in smoosh bash "bash -o posix" dash zsh "zsh --emulate sh" osh mksh ksh yash "yash -o posix"
do
    $SH -c 'echo ${foobar?}' >/dev/null 2>&1
    ec_dashc=$?

    echo 'echo ${foobar?}' | $SH $SH_FLAGS  >/dev/null 2>&1
    ec_stdin=$?

    $SH $SH_FLAGS $tmp >/dev/null 2>&1
    ec_file=$?

    printf "%${WIDTH}s: -c %3d stdin %3d file %3d\n" "$SH" "$ec_dashc" "$ec_stdin" "$ec_file"
done
```
<details>

angelhof:
This is not relevant at the moment (if it does become relevant in the future we could bring this PR back)

PR #333, opened 2021-09-28
Also install python which is needed as a custom aggregator dependency

angelhof:
Custom aggregators require python2 and that is why the `ann_agg` test fails. Eventually we should fix their dependencies to be for python3, but for now we can just download python2 as a dependency before installing PaSh.


angelhof:
@dkarnikis Hmm, it seems that the CI is not working. Could you investigate that? :)

dkarnikis:
@angelhof The curent [configuration](https://github.com/binpash/pash/blob/main/.github/workflows/main.yml#L7) for CI only executes when a file is changed in one or more of the following directories:

 * compiler/**
 * runtime/**
 * evaluation/**
 * annotations/**
 
So it doesn't start spamming tests whenever an irrelevant file is changed.
(I think that if you change a file in on of the directories above the CI should trigger)

angelhof:
Thanks a ton :)

PR #347, opened 2021-10-05
Reduce the number of calls related to distro discovery

dkarnikis:
Added a local config file `~/.pash_init` that currently holds the distro information (and potentially other data we could add in the future). This file is sourced once during the bootstrap of `compiler/pash_init_setup.sh`

### DIFF
--- a/compiler/pash_ptempfile_name.sh
+++ b/compiler/pash_ptempfile_name.sh
@@ -2,13 +2,5 @@
 
 distro=${1??Distro not given}
 # now do different things depending on distro
-case "$distro" in
-    freebsd*)  
-        # bsd is so fun :)
-        tmp=$(TMPDIR=$PASH_TMP_PREFIX mktemp -t pash_XXXXXXXXXX)
-        echo "${tmp}"
-        ;;
-    *)
-        mktemp --tmpdir="$PASH_TMP_PREFIX" -u pash_XXXXXXXXXX
-        ;;
-esac
+tmp_file=$(TMPDIR=$PASH_TMP_PREFIX mktemp -t pash_XXXXXXXXXX)
+echo "${tmp_file}"
--- a/compiler/pash_ptempfile_name.sh
+++ b/compiler/pash_ptempfile_name.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
 distro=${1??Distro not given}
-TMPDIR=$PASH_TMP_PREFIX mktemp -u -t pash_XXXXXXXXXX
+mktemp -u ${PASH_TMP_PREFIX}/pash_XXXXXXXXXX
--- a/compiler/pash_ptempfile_name.sh
+++ b/compiler/pash_ptempfile_name.sh
@@ -1,6 +1,4 @@
 #!/usr/bin/env bash
 
 distro=${1??Distro not given}
-# now do different things depending on distro
-tmp_file=$(TMPDIR=$PASH_TMP_PREFIX mktemp -t pash_XXXXXXXXXX)
-echo "${tmp_file}"
+TMPDIR=$PASH_TMP_PREFIX mktemp -u -t pash_XXXXXXXXXX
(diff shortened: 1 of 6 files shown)


runtime/auto-split.sh
```diff
@@ -8,14 +8,7 @@ n_outputs="$#"
...
 
-if type lsb_release >/dev/null 2>&1 ; then
-   distro=$(lsb_release -i -s)
-elif [ -e /etc/os-release ] ; then
-   distro=$(awk -F= '$1 == "ID" {print $2}' /etc/os-release)
-fi
-
-# convert to lowercase
-distro=$(printf '%s\n' "$distro" | LC_ALL=C tr '[:upper:]' '[:lower:]')
+distro=${1??Distro not given}
```
angelhof, line 11: This change requires that calls to `auto-split.sh` are also modified which would require some changes in the compiler. Maybe we can revert it for now and leave a todo or add an issue?

angelhof (changes requested):
Also maybe I have asked that before but could we maybe do both `freebsd` and other distros using the `freebsd` way (https://github.com/binpash/pash/blob/future/compiler/pash_ptempfile_name.sh#L8)? That would  free us from the need to find the distro in most of these places :)

Furthermore, is https://github.com/binpash/pash/blob/future/compiler/pash_ptempfile_name.sh#L8 the same as https://github.com/binpash/pash/blob/future/compiler/pash_ptempfile_name.sh#L12? I am asking because the second uses the `-u` flag which means `--dry-run` therefore it doesn't create any file :)

runtime/auto-split.sh
dkarnikis: Yeah you are correct, I thought I had addressed this issue.
I am switching all the `mktemp` commands to the freebsd way.
This approach however cannot be applied to all the commands (since the arguments and the flags differ) i.e., https://github.com/binpash/pash/blob/68c5065ef54263806d3bbb37c0ada6d005c76f05/runtime/wait_for_output_and_sigpipe_rest.sh#L15-L23

compiler/pash_ptempfile_name.sh
```diff
@@ -2,13 +2,5 @@
...
-    freebsd*)  
-        # bsd is so fun :)
-        tmp=$(TMPDIR=$PASH_TMP_PREFIX mktemp -t pash_XXXXXXXXXX)
-        echo "${tmp}"
-        ;;
-    *)
-        mktemp --tmpdir="$PASH_TMP_PREFIX" -u pash_XXXXXXXXXX
-        ;;
-esac
+tmp_file=$(TMPDIR=$PASH_TMP_PREFIX mktemp -t pash_XXXXXXXXXX)
```
angelhof, line 5: That is still not the dry-run version. We want the dry-run version in both, not the original one.

compiler/pash_ptempfile_name.sh
```diff
@@ -2,13 +2,5 @@
...
+echo "${tmp_file}"
```
angelhof, line 6: Also this is redundant, we can simply run the above command and its output will be send to stdout. Now we capture the commands output in a variable and then send it to output through stdout :)

runtime/auto-split.sh
```diff
@@ -7,7 +7,7 @@ n_outputs="$#"
 
 # Set a default DISH_TOP in this directory if it doesn't exist
 PASH_TOP=${PASH_TOP:-$(git rev-parse --show-toplevel)}
-
+# TODO change all auto-split calls to provide $distro as argument
```
angelhof, line 10: This TODO is no longer necessary here (also the distro finding) since we don't have a case expression below

angelhof (changes requested):

angelhof:
No worries, let me know when it is done and ready for another review

dkarnikis:
I think its ready for review

angelhof (approved):

PR #349, opened 2021-10-05
Add C++ aggregator for `sort`

janekb04:
I started implementing the C++ aggregator for FreeBSD that corresponds to `sort`. It doesn't support any flags and doesn't pass the unit test yet.


janekb04:
I also realized that for some reason `libdash` updated by itself. I'm not sure if this is wanted or not.

runtime/agg/cpp/aggregators/sort/agg-bsd.h
```diff
@@ -0,0 +1,57 @@
...
+            output() << input1_top << '\n';
+            refill1 = true;
+        }
+        else
+        {
+            output() << input2_top << '\n';
+            refill2 = true;
+        }
+    }
+
```
janekb04, line 51: `if(!refill1) output << input1_top;`

janekb04:
@nvasilakis Now the simple `sort` aggregator works. It passes the unit test. It is basically the _merge_ step from _mergesort_. Implementing support for `sort` flags should be straightforward, but will take some time due to their amount. I plan to do this using either higher-order functions or polymorphism. I will see what will be easier and more performant.

janekb04:
I updated the main comment with some notes and an option to-do list for better organization.

nvasilakis:
Hi @janekb04 are you planning on completing this so that we merge it into mainline?

PR #350, opened 2021-10-06
Python optimizations to reduce startup time [WIP]

dkarnikis:
Switching to python -S flag.
Added local directory to download the required python packages and export PYTHONPATH to point to it (in pa.sh).
Removed unused imports from all the python files.


dkarnikis:
Executing the interfaces benchmark suite, we get a 5.49% (~150ms) faster execution with the current changes on a
Intel(R) Core(TM) i5-10210U CPU @ 1.60GHz.

dkarnikis:
With the json change, running again the interface benchmark suite, we get:
14.86% faster execution (~500ms).
Testing these changes when executing an empty file, we get:
30.98% faster execution (~0.020ms)

PR #354, opened 2021-10-08
switch Arch installation to PKGBUILD

dkarnikis:
Remove arch installation commands from distro-deps.sh
Added arch PKGBUILD in the repo (Should do the same .deb and other packages)


dkarnikis:
Fixed in https://github.com/binpash/pash/pull/484

PR #360, opened 2021-10-10
PaSh compiler expands by querying a bash process

angelhof:
Instead of using and maintaining our own expansion in python, we can simply ask a bash to expand strings for us, duh!

The current design is as follows. The daemon spawns a bash process that will mirror the variables (and other state) of the original bash and then queries that process to expand strings for it. Since querying happens without bash startup it should be pretty fast.

An alternative design would be for the daemon to have a `communication` with the original shell when compiling, asking it for more info, like variable values etc. This would also reduce the overhead of variable passing which happens now always.

I think I prefer the alternative design. The concern is that this might interfere with the speculative parallelism that @tammam1998 is working on. Tammam what do you think? It would also require some additions to the runtime daemon client to have a multi-round communication with the daemon.

@mgree Should we replace python expansion with that? Do you see any issues?

@nvasilakis what do you think?

### DIFF
--- a/compiler/expand.py
+++ b/compiler/expand.py
@@ -282,7 +282,7 @@ def lookup_variable_inner(varname):
     if config.pash_args.expand_using_bash_mirror:
         return config.query_expand_variable_bash_mirror(varname)
     else:
-        _type, value = config.config['shell_variables'][varname]
+        _type, value = config.config['shell_variables'].get(varname, [None, None])
         return value
 
 def invalidate_variable(var, reason, config):


mgree:
Expansion has side-effects, e.g., `${x=foo}`. This approach should work, but we'd have to check to see whether or not some variables got updated.

I think the only forms that mutate the actual shell environment are `${x=...}`, `${x:=...}`, and arithmetic forms like `$(( x+= 1 ))` or `$(( x++ ))` (POSIX optional, but not uncommon).

PR #374, opened 2021-11-08
Add buffer to scheduler to deal with multiple inputs

tammam1998:


compiler/pash_runtime_daemon.py
```diff
@@ -225,11 +231,13 @@ def parse_and_run_cmd(self, input_cmd):
...
 
     def wait_input(self):
-        input_cmd = ""
-        with open(self.in_filename) as fin:
-            input_cmd = fin.read()
-            input_cmd = input_cmd.rstrip()
-        return input_cmd
+        cmd = ""
+        if not self.reader_pipes_are_blocking:
+            while not cmd:
```
angelhof, line 236: We should move this while internally in get_next_cmd(). The scheduler should not be exposed to those things

compiler/pash_runtime_daemon.py
```diff
@@ -262,10 +270,53 @@ def run(self, in_filename, out_filename):
...
+        self.reader.close()
         shutdown()
 
 
+class UnixPipeReader:
+    def __init__(self, in_filename, blocking = True):
+        self.in_filename = in_filename
+        self.buffer = ""
+        self.blocking = blocking
+        if not self.blocking:
```
angelhof, line 283: Log whether the initialization was blocking or non-blocking (not that we will use the non-blocking API).

angelhof: Actually, we should add a comment that non-blocking is not supposed to be used. And that it was just an experiment.

compiler/pash_runtime_daemon.py
```diff
@@ -262,10 +270,53 @@ def run(self, in_filename, out_filename):
...
+            self.fin = open(self.in_filename)
+
+    def get_next_cmd(self):
```
angelhof, line 286: We should add a comment explaining what this method does. What can the possible starting states be (e.g., buffer can contain complete commands, at least one, no more than X, etc). And then what happens when the get_next_cmd is done. What is returned and what should we expect from the state.

compiler/pash_runtime_daemon.py
```diff
@@ -262,10 +270,53 @@ def run(self, in_filename, out_filename):
...
+        log("Previous cmd buffer:", self.buffer)
+        log("Previous cmd buffer length:", len(self.buffer))
+        input_buffer = ""
+        if self.buffer:
+            # Don't wait on fin if cmd buffer isn't empty
+            input_buffer = self.buffer
```
angelhof, line 292: We should log that we are using the previous buffer here, instead of reading for new inputs.

compiler/pash_runtime_daemon.py
```diff
@@ -262,10 +270,53 @@ def run(self, in_filename, out_filename):
...
+class UnixPipeReader:
+    def __init__(self, in_filename, blocking = True):
+        self.in_filename = in_filename
+        self.buffer = ""
+        self.blocking = blocking
+        if not self.blocking:
+            self.fin = open(self.in_filename)
+
+    def get_next_cmd(self):
+        log("Previous cmd buffer:", self.buffer)
```
angelhof, line 287: I would remove this log and the next.

compiler/pash_runtime_daemon.py
```diff
@@ -262,10 +270,53 @@ def run(self, in_filename, out_filename):
...
+            self.buffer = rest
+        else:
+            cmd = input_buffer
+            self.buffer = ""
+
+        cmd = cmd.rstrip()
+        log("Reader returned cmd:", cmd)
+        return cmd
+
+    def close(self):
```
angelhof, line 316: we could log that the reader closes.

compiler/pash_runtime_daemon.py
tammam1998: I think it makes sense to have this in the scheduler because the scheduler initializes the UnixPipeReader as nonblocking so it should deal with the behavior of nonblocking reading.

angelhof:
There is still a rarely occuring synchronization bug which will be solved later in another PR.

PR #384, opened 2021-11-15
Exposing the exit status of pash-compiled pipelines

angelhof:
This PR tries to add support for returning the exit status of pash-compiled pipelines. These changes fix the issue in general but there is still a failure with `--parallel_pipelines` since this flag executes pipelines in the background and therefore their exit status is lost.

@mgree, @tammam1998 the problem is here: https://github.com/binpash/pash/blob/cat-redir-fail/compiler/pash_runtime.sh#L233. Is there any way to address this without losing support for parallel pipelines?

### DIFF
--- a/compiler/ast_to_ir.py
+++ b/compiler/ast_to_ir.py
@@ -45,34 +45,34 @@ compile_cases = {
         }
 
 preprocess_cases = {
-    "Pipe": (lambda irFileGen, config:
-             lambda ast_node: preprocess_node_pipe(ast_node, irFileGen, config)),
-    "Command": (lambda irFileGen, config:
-                lambda ast_node: preprocess_node_command(ast_node, irFileGen, config)),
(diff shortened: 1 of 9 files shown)


angelhof:
Another thought: Maybe we have a check that disables `parallel-pipelines` in the context of `&&`, `||`, `[ ]`. This does not address the issue with expanding `$?` though. Also, the final dataflow region can also have the `parallel-pipelines disabled`.

angelhof:
Also `set -e`.

angelhof:
@tammam1998, @dkarnikis This change might affect the performance of `for-loop` benchmarks with parallel pipelines since it disables the `parallel_pipelines` optimization for the final command of a script, since we might need its exit code. It can be easily addressed by concluding scripts with a final command, e.g., `echo done`, to make sure that the `for` loops will execute in parallel.

Issue #381, opened 2021-11-15
Pash not running on Debian


dkarnikis:
Executing any command with PaSh hangs during execution.

Reproduce:
`
$PASH_TOP/pa.sh -c 'echo test'
`

will neither print nor close PaSh.
Even running `$PASH_TOP/pa.sh --help` will not exit (however, it will print the usage log message).
By setting `pash_daemon=0` in pa.sh seems to be a temp fix (for me at least).

Commit: ea6c3a642c131094436b34d57e3719c66b9489d6
Branch: future
Kernel:  Linux livestar 4.9.0-16-amd64 #1 SMP Debian 4.9.272-1 (2021-06-21) x86_64 GNU/Linux
OS: Debian GNU/Linux 9 (stretch)
Machine: Livestar

angelhof:
This probably has to do with the latest changes. Now the pash `daemon` communicates through sockets. I suspect that this is either a problem with some usage of `nc` (in `compiler/pash_init_setup.sh`) or the inability of the daemon to bind and listen on the socket. Could you use `-d 1` and tell me what happens?

BTW, using `--no_daemon` flag will have the same behavior as setting that flag manually.

dkarnikis:
```sh
│Daemon:  WARNING: Could not parse annotation for file: /home/dkarnikis/fpash/annotations/pr.json
│Daemon:  |-- Expecting property name enclosed in double quotes: line 14 column 18 (char 285)
│Daemon:  SocketManager: Made sure that socket does not exist
│Daemon:  SocketManager: Created socket
│Daemon:  SocketManager: Successfully bound to socket
│Daemon:  SocketManager: Listenting on socket
│Daemon:  Received data:
│Daemon:  Received data:
│Daemon:  Received data:
```
and several more `Daemon:  Received data:`

angelhof:
It seems to me then that this might be a problem with `nc`. 

This function https://github.com/binpash/pash/blob/future/compiler/pash_init_setup.sh#L185 is called in `pa.sh`. Could you modify it to not send it stdout and stderr to `/dev/null` and instead print it out for debugging?

dkarnikis:
```sh
nc: unix connect failed: No such file or directory
nc: unix connect failed: No such file or directory
nc: unix connect failed: No such file or directory
nc: unix connect failed: No such file or directory
nc: unix connect failed: No such file or directory
nc: unix connect failed: No such file or directory
nc: unix connect failed: No such file or directory
[...]

PR #403, opened 2021-11-23
Livestar fixes

angelhof:
This PR contains fixes or changes so that pash works on livestar.

1. It modifies `nc -z` to not use that flag to check if daemon is listening since for some reason, the `nc` version on livestar always returns 1 even if it is connected.
2. Modifies test input for `spell-grep` for tests to take reasonable time.

### DIFF
--- a/compiler/pash_init_setup.sh
+++ b/compiler/pash_init_setup.sh
@@ -195,7 +195,7 @@ else
         maximum_retries=100
         ## For some reason, `nc -z` doesn't work on livestar (it always returns error)
         ## and therefore we need to send something. 
-        until  echo "Daemon Start" | nc -U "$DAEMON_SOCKET" >/dev/null 2>&1 ; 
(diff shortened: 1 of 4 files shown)


angelhof:
I have absolutely no clue what fails in Ubuntu 20 :')

dkarnikis:
The good thing is that you fixed this issue https://github.com/binpash/pash/issues/381. 
However, the problem still persists for me (and checking with htop, your tests has also stopped at 20.sh)

angelhof:
Cool, I will investigate why it hangs (and also what breaks in Ubuntu 20 )

angelhof:
Everything seems to be good here! @dkarnikis if this fixes hanging issues with `20.sh` let's merge!

dkarnikis:
Alright, I will check it in a bit and I will merge it! Thanks

dkarnikis:
20.sh runs and prints the results! Well done!
Merging to future

PR #411, opened 2021-11-29
Fix an issue with diff annotation

angelhof:
@nvasilakis  Can you think of a reason why we had `diff` be non-pure? I don't see why `diff` is not pure by default.

@dkarnikis after/if we merge this we should rerun `oneliners/diff.sh`


angelhof:
Let's merge this after Tammam's PR!

PR #415, opened 2021-12-01
Added logic to make sure to improve performance when input is small

tammam1998:


angelhof:
Let's run all tests with the experimental flags on deathstar before merging. @dkarnikis can you do it?

dkarnikis:
@angelhof do we need the parallel flag as well?

PR #417, opened 2021-12-03
Profile driven width lowering

angelhof:
This PR introduces profile-driven width lowering. If the same part of a script is recompiled, the daemon determines its width based on past execution times.


dkarnikis:
Merging to future

PR #419, opened 2021-12-05
Gather pids to kill when spawning

angelhof:
Instead of finding which pids we need to stop afterwards, we now do it while spawning them. This should improve performance.

### DIFF
--- a/compiler/config.py
+++ b/compiler/config.py
@@ -301,6 +301,8 @@ def read_vars_file(var_file_path):
             print_time_delta("Variable Reading", variable_reading_start_time, variable_reading_end_time)
 
             variable_tokenizing_start_time = datetime.now()
+            ## TODO: Can we replace this tokenizing process with our own code? This is very slow :'(
+            ##       It takes about 15ms on deathstar.
             tokens = shlex.split(data)
             variable_tokenizing_end_time = datetime.now()
             print_time_delta("Variable Tokenizing", variable_tokenizing_start_time, variable_tokenizing_end_time)
--- a/compiler/config.py
+++ b/compiler/config.py
@@ -4,6 +4,8 @@ import subprocess
 import math
 import shlex
 
+from datetime import datetime
+
 from ir_utils import *
 from util import *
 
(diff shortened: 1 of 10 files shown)


compiler/config.py
```diff
@@ -293,8 +295,17 @@ def read_vars_file(var_file_path):
...
         with open(var_file_path) as f:
+            variable_reading_start_time = datetime.now()
             data = f.read()
+            variable_reading_end_time = datetime.now()
+            print_time_delta("Variable Reading", variable_reading_start_time, variable_reading_end_time)
+
+            variable_tokenizing_start_time = datetime.now()
+            ## TODO: Can we replace this tokenizing process with our own code? This is very slow :'(
+            ##       It takes about 15ms on deathstar.
             tokens = shlex.split(data)
```
angelhof, line 306: @mgree Do you have any thoughts (or would like to try) to implement an custom/alternative tokenization? In the past we had our own but it didn't work for new lines. Using python's `shlex` works, but it is extremely slow (it takes about 15ms on deathstar). This is a cost we pay everytime we try to compile a script fragment, and currently is the biggest overhead in the runtime.

mgree: For context... this is variable saving/loading?

My first move would be to us shlex more carefully: [every call to `shlex.split` instantiates a new `shlex.shlex` object](https://github.com/python/cpython/blob/3.10/Lib/shlex.py#L305). Could we reuse them?

After that... maybe we could save a bit of time by working lazily, rather than [materializing the token list](https://github.com/python/cpython/blob/cb68c0a3a4aeb4ec58ab1f71b70bc8bfecbceef6/Lib/shlex.py#L315)?

After _that_... take a look at [bashlex's tokenizer](https://github.com/idank/bashlex/blob/master/bashlex/tokenizer.py)? I'd be hesitant to write my own tokenizer, since it'll be hard to get it exactly right, but custom code in C would probably be the fastest option.

angelhof: I already tried to reuse the object but I couldn't make it work. I then searched online and found several threads that `shlex` is known to be very slow.

As far as I understand laziness also shouldn't make a significant difference since that list isn't that big anyway.

I will try to take a look at bashlex's tokenizer, though I don't think that is what we want. I think tokenization is the wrong word actually, because we just want to parse whitespace separated words, correctly taking into account quotes and newlines.

mgree: Ah. Got it. These are variable definitions... no functions? If there are no functions, it shouldn't be too bad at all to roll something custom.

angelhof: Yes that is right, there are only variable definitions!

mgree: Okay, cool. I'm happy to take a whack at this... after the deadline?

PR #423, opened 2021-12-09
Pure expansion

mgree:
The expansion routine was expanding assignments, which is unsound with the current pipelining approach.

This commit turns it off, with a new `ImpureExpansion` exception raised when the `config` could be updated.


angelhof:
Someone (probably me) should run the POSIX tests with this PR and if it works (which it should) we should just merge it.

angelhof:
TODO (for me): Add a test that exercises the behavior in this PR, run tests, and merge.

Issue #434, opened 2022-01-14
Installation issues.


valuta1995:
While following the instructions on docker from source I came across a few problems.

- On the line `docker build -t "pash/18.04" .` I need to add the `-f` flag with the path to the Dockerfile (`./docker/ubuntu/Dockerfile`)
- There is no `/pash` in the docker container, there however is `/opt/pash`
- pa.sh does not run in the container as it cannot find pexpect. The workaround I used is to remove the `usr/local/lib/python3.6/dist-packages` folders from python_pkgs and place their content in the python_pkgs folder.
- Before the previous step and after, running pa.sh gives the following error:

```
Error: Maximum retries: 100 exceeded when waiting for daemon to bind to socket!
```

valuta1995:
I made two pull requests for the doc changes. I will not touch the folder structure as I'm unsure as to how to do that properly (without messing other parts up).

dkarnikis:
Hey @valuta1995 . Thank you for your feedback and your PRs. Could you try signing your commits with: 
`git commit -s -m 'git message'` to pass the DCO test?
You may find additional information [here](https://github.com/binpash/pash/pull/435/checks?check_run_id=4817077894) for the #435 PR

valuta1995:
I used the editor on github, which does not play well with DCO, so I made a new request through cli.

angelhof:
Hi @valuta1995! Thanks a lot for letting us know of these issues. Me and @dkarnikis will try to debug maximum retry error as soon as possible. We are drowning in work though so it might take a few days!

angelhof:
This PR #439 should address the maximum retries issue. We will try to automatically fix the python_pkg issue in the next few days but for now you should be good to go.

valuta1995:
To quickly verify I installed netcat-openbsd (`sudo apt install netcat-openbsd`) manually in the container. This worked. Thank you for your help!

dkarnikis:
#440 Should fix the python_pkg issue. Could you verify if it works for you as well?

PR #439, opened 2022-01-14
Install netcat-openbsd 

angelhof:
PaSh needs the openbsd version of `netcat` since it uses it for unix-socket communication. This addresses part of the issue in: #434


angelhof:
@dkarnikis if this PR looks good let's just merge it!

dkarnikis:
@angelhof could we delete this PR since it is included in https://github.com/binpash/pash/pull/445 ?

angelhof:
I don't see why. We can merge this, then do a merge of main on top of the other, and then merge the other on main too.

PR #440, opened 2022-01-18
Fix for python pkgs in pash setup

dkarnikis:
This PR should address https://github.com/binpash/pash/issues/434


angelhof (approved):
Looks good to me. If the change indeed fixes the docker installation, then we should proceed!

PR #442, opened 2022-01-20
Fixes

dkarnikis:
General bug-fixes


compiler/test_evaluation_scripts.sh
```diff
@@ -171,15 +171,14 @@ execute_tests() {
         for conf in "${configurations[@]}"; do
             for n_in in "${n_inputs[@]}"; do
                 echo "|-- Executing with pash --width ${n_in} ${conf}..."
-                export pash_time="${test_results_dir}/${microbenchmark}_${n_in}_distr_${conf}.time"
+                export pash_time="${test_results_dir}/${microbenchmark}_${n_in}_distr_$(echo ${conf} | tr -d ' ').time"
```
angelhof, line 174: why is this needed?

compiler/util.py
```diff
@@ -19,7 +19,11 @@ def pad(lst, index):
 def print_time_delta(prefix, start_time, end_time, args=None):
     ## Always output time in the log.
     time_difference = (end_time - start_time) / timedelta(milliseconds=1)
-    log("{} time:".format(prefix), time_difference, " ms")
+    ## If output_time flag is set, log the time
```
angelhof, line 22: Just to clarify, this logs the time regardless of the value of `-d` if we have `--output_time` enabled, right? I think that is the correct behavior!

angelhof: If it is, we probably need to be a bit more clear about it in the comment.

evaluation/benchmarks/run.seq.sh
```diff
@@ -52,7 +52,11 @@ oneliners(){
     IFS=";" read -r -a script_input_parsed <<< "${script_input}"
     script="${script_input_parsed[0]}"
     input="${script_input_parsed[1]}"
-    export IN="$PASH_TOP/evaluation/benchmarks/oneliners/input/$input"
+    if [[ "$1" == "--small" ]]; then
```
angelhof, line 55: It is a little ugly that this needs to be duplicated in both run.seq.sh and run.par.sh. Could you think of a way to abstract this behavior (and many others that are the same) in a common function/file that both run.par.sh and run.seq.sh use?

compiler/test_evaluation_scripts.sh
angelhof: ah good point, thanks!

compiler/util.py
dkarnikis: That is correct. When using `--output_time`, we get the time for each operation without the need of the `-d` flag!

evaluation/benchmarks/run.seq.sh
dkarnikis: That's quite a good point. I could create a new function that will be called from each benchmark during its startup. Moreover, I am thinking of moving the `export`- ing of the respective variables (`IN IN_PRE` etc to the `setup.sh`). What do you think?

angelhof: My only issue with moving exporting to setup is that one might want to setup different benchmark suites all together in the beginning and then run them one by one. This would break the flow. I think that exporting of respective variables should happen as close as possible to the actual execution (maybe it can be a function that is defined in setup.sh and then sourced, but exporting should not be tied with downloading inputs IMO.

dkarnikis:
There is a strange bug with temp-analytics that has been annoying me the past few hours and I cannot comprehend what causes it (I think it's irrelevant with the previous issue).

dkarnikis:
@angelhof feel free to review it again (no rush)!

evaluation/benchmarks/analytics-mts/input/setup.sh
```diff
@@ -13,13 +13,26 @@ if [[ "$1" == "-c" ]]; then
...
+    "$PASH_TOP/scripts/append_nl_if_not.sh"  in.csv
+    len=$(cat in.csv | wc -l)
+    half_size=$(( $len / 4 ))
+    head -n $half_size in.csv > in_small.csv
   fi
-  "$PASH_TOP/scripts/append_nl_if_not.sh"  in.csv
-fi
+}
+
+source_var() {
```
angelhof, line 32: So what is the plan for using it? we need to `source setup.sh` and then run the actual test?

evaluation/benchmarks/run.seq.sh
```diff
@@ -143,11 +140,11 @@ max-temp(){
   cd max-temp/
 
   mkdir -p "$outputs_dir"
-
+  export IN=
```
angelhof, line 143: I am confused because in some benchmarks we get the variable from `setup.sh` and then `source_var`, and in other we export it here.

scripts/utils.sh
```diff
@@ -43,3 +43,19 @@ append_nl_if_not(){
   fi
 }
 
+download_dataset() {
```
angelhof, line 46: I find this name confusing because it doesn't just download the dataset, but also installs dependencies and sources the setup. I would call this function `install_deps_source_setup` or something more descriptive like that :)

angelhof (approved):
Everything looks good modulo some renamings that I think would help understanding.

evaluation/benchmarks/analytics-mts/input/setup.sh
angelhof: I think that is OK! My confusion was resolved when I saw the `download_...` function in `utils.sh`

evaluation/benchmarks/run.seq.sh
angelhof: I saw it in one more place, I think `oneliners`.

angelhof:
This PR is huge. @dkarnikis and @nvasilakis if tests pass and it is an adequate checkpoint here, let's merge!

dkarnikis:
@angelhof we should squash and merge right?

angelhof:
> @angelhof we should squash and merge right?

Ideally yes. Except if you think that this PR contains a few big chunks of work, where you could rebase it into a few (<5) commits with clean messages and then just merge.

dkarnikis:
I prefer not to mess with it anymore.
Merging to future, I hope nothing will break :)

Issue #438, opened 2022-01-14
Potential redundancy in Dockerfile and distro-deps.sh


angelhof:
Do we need both the separate dockerfiles for each distro, and the case analysis in distro-deps? Could we get away with having one Dockerfile and only checking in `distro-deps.sh`?

dkarnikis:
I think the case-analysis in distro-deps is required  whether we are using docker or not.
Moreover, these images are prebuilt and available [`here`](https://hub.docker.com/repository/docker/binpash/pash).

So, I suppose we could try removing(?)  the dockerfiles for debian and fedora from the repo, and updating the correct documents. Should I proceed with that?

angelhof:
That sounds good. If they are indeed not useful, we could try to delete the unnecessary dockerfiles. But probably this is not a priority.

dkarnikis:
In [dockerhub](https://hub.docker.com/r/binpash/pash), we have several references to these dockerfiles in the repo.  Should I remove these as wel?

angelhof:
I think I misexpressed my point. My point was simply that both `distro-deps.sh` and the dockerfiles, have partly the same functionality. So I am confused why we need both. Our dockerfiles only call some shell commands, so couldn't they just use `distro-deps.sh`? And then we can have 3 dockerfiles, but all of them just call distro-deps and do nothing else.

PR #445, opened 2022-01-25
Dockerfiles fix

dkarnikis:
- Fixed dockerfiles for ubuntu, debian and fedora.
- Moved all the deps inside `distro-deps.sh`
- Removed sudo from original installation scripts

`git` is the only package required in the dockerfiles (in order to fetch the repo).
This PR includes https://github.com/binpash/pash/pull/439 and should fix https://github.com/binpash/pash/issues/438

Signed-off-by: dkarnikis <dkarnikis>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,13 +19,15 @@ jobs:
       runs-on: ${{ matrix.os }}
       if: github.event.pull_request.draft == false
       steps:
+            
         - uses: actions/checkout@v2 
           with:
             ref: ${{ github.event.pull_request.head.sha }}
         - name: Running-Pash-Benchmarks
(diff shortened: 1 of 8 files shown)


angelhof (approved):
Thanks! Looks good to me! The dockerfiles are now much cleaner. Before merging, I would suggest following the installation process on a new clean machine (could also be a VM/docker container) at least for Ubuntu so that we know that nothing broke.

dkarnikis:
So all the tests have passed. I also deployed a fresh installation of pash on an ubuntu 20.04 and had no issue as well (all the tests passed too). @angelhof should we merge?

angelhof:
Great job! Yes, we should merge :)

Issue #443, opened 2022-01-20
Temp Analytics error


dkarnikis:
When running temp-analytics.sh with `--dgsh_tee --parallel_pipelines --profile_driven` (blish_no_comm), 
I get the following error on deathstar:
```sh
executing max temp with pash Thu 20 Jan 2022 07:15:35 AM EST
temp-analytics.sh:                                          
gzip: stdin: invalid compressed data--format violated       
(23) Failed writing body 
(23) Failed writing body 
(23) Failed writing body                                    
....
(23) Failed writing body

gzip: stdin: invalid compressed data--format violated
(23) Failed writing body
(23) Failed writing body
(23) Failed writing body
(23) Failed writing body
874.265
```
I don't have the same issue when running other flag configurations.
Running on branch future

angelhof:
Could you try rerunning it with all subsets of 2 out of 3 of these flags and see if it fails for those too? Also running it with the addition of `--r_split` too? I am not sure why this would be an issue with this exact flag combination.

dkarnikis:
Running with the following configurations, I think the bug is related to `r_split`:
```
export PASH_FLAGS="--r_split --dgsh_tee --parallel_pipelines --profile_driven"
421.258
export PASH_FLAGS="--r_split --dgsh_tee --parallel_pipelines"
428.272
export PASH_FLAGS="--r_split --dgsh_tee"
616.152
export PASH_FLAGS="--dgsh_tee"
error
export PASH_FLAGS="--dgsh_tee --parallel_pipelines --profile_driven"
error
export PASH_FLAGS="--dgsh_tee --parallel_pipelines"
error
export PASH_FLAGS=" "
1278
```

angelhof:
Hmm, so it seems to be that having dgsh-tee without r-split creates a problem. @tammam1998 could you investigate this?

PR #447, opened 2022-01-26
Evaluation script

dkarnikis:
Script to gather execution results for:
- oneliners
- unix50
- analytics-mts
- poets
- max-temp
- web-index
- dependency_untangling

using several configurations


dkarnikis:
After we fix https://github.com/binpash/pash/issues/443, we could merge the PR

angelhof:
That is great! If I understand correctly it runs all benchmarks with all configurations once. However, if I recall correctly, in our evaluation we run all configurations of all benchmarks 3 times. Is it possible to extend this script to do that?

dkarnikis:
> That is great! If I understand correctly it runs all benchmarks with all configurations once. However, if I recall correctly, in our evaluation we run all configurations of all benchmarks 3 times. Is it possible to extend this script to do that?

I will work on that! 
TODO: find a way to merge the results in order to produce the final plots. Should I add the scripts for the plot generation here?

angelhof:
Great :) @mgree do you think we could automate using the R scripts or should we just have a more ad-hoc infrastructure (e.g., Python pyplot) for CI and artifact (and only use the R scripts for the final plots once).

mgree:
I'd say we might as well use the R plots, just so it's always the right thing. There's an example of how to do it in [CI for ffs](https://github.com/mgree/ffs/blob/main/.github/workflows/build.yml). Let me know if I can help!

nvasilakis:
@dkarnikis Please add README.md on how to run the full artifact and get the results (see link from previous artifacts).

@tammam1998 Please run this infrastructure to confirm we can get the results from the paper.

@angelhof We need to decide how to call the "legacy" configuration, and how to run/incorporate the POSIX or Smoosh evaluation.

angelhof:
> @angelhof We need to decide how to call the "legacy" configuration, and how to run/incorporate the POSIX or Smoosh evaluation.

Regarding the legacy configurations in the artifact, I think we should use the names in the paper and define them in a README here.

Regarding the POSIX evaluation, I think it is totally fine to not give access to it, or, at most, give them an ssh key to run the tests on our machine and check the git commit hash to ensure that it is correct.

nvasilakis:
But the name we're using is only for blinding purposes—we will eventually use a different name at the end, so why not use that final name in these scripts?:-)

nvasilakis:
> But the name we're using is only for blinding purposes—we will eventually use a different name at the end, so why not use that final name in these scripts?:-) ( ...so I'm asking for your opinion on what that name should be:-) )

@dkarnikis let's call the legacy setup `pash-aot` (with proper capitalization, depending on where it's used.) Thanks, then we can merge, and @tammam1998  can test asynchronously.

nvasilakis:
I think @tammam1998 has tested and we can merge.

nvasilakis (approved):
Thank you! Let's merge, if @angelhof agrees.

PR #455, opened 2022-02-05
fix bug caused by early kill of process

tammam1998:
This is a rare bug that happens sometimes when some processes terminate before the sigterm signal reaches other processes. This bug happens non deterministically (only sometimes) on 20.sh and 34.sh with width 16.


angelhof (approved):
Looks good to me!

PR #456, opened 2022-02-08
PaSh citation in README

angelhof:
I think it makes sense to have the citation (and/or the main paper) accessible somewhere in the README. 

I wrote a stub, which we can improve to add more info or modify the text.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 ## PaSh: Light-touch Data-Parallel Shell Processing
 > _A system for parallelizing POSIX shell scripts._
 
-Quick Jump: [Running PaSh](#running-pash) | [Installation](#installation) | [Testing](#testing) | [Repo Structure](#repo-structure) | [Community & More](#community--more)
+Quick Jump: [Running PaSh](#running-pash) | [Installation](#installation) | [Testing](#testing) | [Repo Structure](#repo-structure) | [Community & More](#community--more) | [Citing](#citing)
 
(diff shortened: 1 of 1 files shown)


nvasilakis:
We have all the relevant papers in [./docs/README](https://github.com/binpash/pash/tree/main/docs#academic-papers--events) and the website (which, ironically, [earlier had bibtex entries too, removed upon "someone's" request](https://github.com/binpash/pash/commit/0dfaaef422548836b2ee8b1e416fec68cf8feacb) hehe:-)

Would it make sene to bring these bibtex entires back there and simply add a pointer into the top-level README?

angelhof:
> We have all the relevant papers in [./docs/README](https://github.com/binpash/pash/tree/main/docs#academic-papers--events) and the website (which, ironically, [earlier had bibtex entries too, removed upon "someone's" request](https://github.com/binpash/pash/commit/0dfaaef422548836b2ee8b1e416fec68cf8feacb) hehe:-)

Was that me? Oops :) Yes, I remember that they are there, but they are difficult to find (and also there is no definitive citation for the project). I was thinking of a citation for the actual PaSh software (that is why I only added the EuroSys paper too).
 
> Would it make sene to bring these bibtex entires back there and simply add a pointer into the top-level README?

I think that would make a lot of sense. We should certainly add a pointer, because the papers are not easy to find in docs. If we think that it makes sense to also have a citation in the main README that is fine too I think.

PR #458, opened 2022-02-14
Kill processes that are exclusively under the current process tree

dkarnikis:
- Kill pending processes running under the current instance of pash (killed by signal)
- Doesn't affect the performance of other running pash instances

(Moved from https://github.com/binpash/pash/pull/454)


nvasilakis (approved):
Looks good, let's merge!

Issue #450, opened 2022-02-02
Add appropriate env vars to .bashrc/.zshrc/... 


nvasilakis:
After the setup script completes, the user is expected to set up the `PASH_TOP` environment variable (and, optionally, update the `PATH` to easily locate `pa.sh`). Can we have the setup script:

1. check the existence of `~/*.rc` shell config files (from a set of common shell config files — for example, I have both `.bashrc` and `.zshrc`)
2. ask the user if they want their `.*rc` automatically updated (could also pass a flag)
3. check that there are no such lines in each `.*rc`: if true, append config lines; if false, replace them.

PR #459, opened 2022-02-14
export PASH_TOP and PATH after pash installation

dkarnikis:
Addresses https://github.com/binpash/pash/issues/450

### DIFF
--- a/scripts/setup-pash.sh
+++ b/scripts/setup-pash.sh
@@ -110,3 +110,28 @@ $PASH_TOP/evaluation/tests/input/setup.sh
 echo " * * * "
 echo "Do not forget to export PASH_TOP before using pash: \`export PASH_TOP=$PASH_TOP\`"
 echo '(optionally, you can update PATH to include it: `export PATH=$PATH:$PASH_TOP`)'
+echo " * * * "
+## append PASH Configuration paths to the respective rc files
+rc_configs=(~/.shrc ~/.bashrc  ~/.zshrc ~/.cshrc ~/.kshrc) # add more shell configs here
+for config in "${rc_configs[@]}"
+do
+    ## if the config exists
+    ## check if it contains an old entry of Pash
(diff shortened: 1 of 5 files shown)


scripts/setup-pash.sh
```diff
@@ -110,3 +110,28 @@ $PASH_TOP/evaluation/tests/input/setup.sh
...
+        # get the shell name
+        shell_name=$(echo $(basename $config) | sed 's/rc//g' | sed 's/\.//g')
+        echo "Do you want to append \$PASH_TOP to $shell_name ($config) (y/n)?"
+        read answer
+        if [ "$answer" != "${answer#[Yy]}" ] ;then 
+            tmpfile=$(mktemp -u /tmp/tmp.XXXXXX)
+            # create a backup of the shell config
+            cp $config ${config}.backup
+            # remove all the entries pointing to PASH_TOP and PATH
+            cat $config | grep -v -e "export PASH_TOP" -e "export PATH" > $tmpfile
```
angelhof, line 130: I think it is very aggressive to remove previous $PATH entries. There is no need at all to remove PATH entries or to expand `$PATH` as done below (L 134). We can simply append a line that does not expand the variables at all:
```
echo 'export PATH=$PATH:$PASH_TOP' >> $config
```
Maybe we could remove previous exports of PATH if and only if they contain $PASH_TOP in them (since removing the PASH_TOP will leave them blank.

Regarding the $PASH_TOP, entry, I guess it is fine to remove it.

angelhof (changes requested):

dkarnikis:
Note, we (I probably) should also fix the CI and docker setup to handle these changes to pass 'y' as input automatically

angelhof (approved):
Looks good to me now!

dkarnikis:
Hope nothing will break with the automated scripts!
Merging

PR #460, opened 2022-02-14
Add file about troubleshooting in docs

festutz:
When setting up PaSh, I encountered some obstacles I could not solve with the current documentation.
@dkarnikis explained how to overcome them. Following advice by him and @nvasilakis, I summarized this as a first item in a list for troubleshooting.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -1,0 +1,14 @@
+# PaSh Troubleshooting
+
+This is a list of issues when setting up or running PaSh (and how to solve them).
+
+### Module `pexpect` not found OR libdash library for parser not working  
+
+_Solution_: run dependency scipt with **sudo** and build pash libraries 
+
+```console
+cd $PASH_TOP  
+sudo bash scripts/distro-deps.sh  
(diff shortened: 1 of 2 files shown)


dkarnikis:
@festutz Thank you for the PR! 
FYI, for every commit in the repo, several tests correctness (and not only tests run). In the current PR, the DCO test failed (it can be easily fixed by checking [here](https://github.com/binpash/pash/pull/460/checks?check_run_id=5183419593)).

To avoid it in the future, you could try signing your commits by adding the `-s` flag when running `git commit`

festutz:
@dkarnikis Thanks! I fixed it accordingly.

nvasilakis:
Thanks @festutz !

We should also update the appropriate READMEs and tutorial so that people don't hit this problem in the first place. And maybe add some checks in `distro-deps.sh` to (1) warn users it needs to be executed using `sudo` and (2) warn users if it hasn't been invoked using `sudo`?

(not sure what's best, because I see that in some platforms — Docker? — it even installs `sudo`)

nvasilakis (approved):

PR #461, opened 2022-02-14
Fix link to guide for contributing

festutz:
The link was broken and since there is only one single file in the directory, I changed it to point it to this one.


dkarnikis:
Seems good to me! 
Merging to me

PR #465, opened 2022-02-20
add unzip dep

dkarnikis:


dkarnikis:
Addressed in https://github.com/binpash/pash/pull/467

PR #468, opened 2022-02-21
update up.sh

dogi:
binpash instead of andromeda


nvasilakis (approved):
Great catch, @dogi, thanks so much!

PR #469, opened 2022-02-21
Fix input file for shortest scripts, i.e., make it all_cmds

festutz:
The original file was not a list of commands and does not really make sense for the script I think. From the input folder, the file `all_cmds.txt` was reasonable to choose but I do not exactly know whether this was the original intention.

### DIFF
--- a/evaluation/benchmarks/oneliners/shortest-scripts.sh
+++ b/evaluation/benchmarks/oneliners/shortest-scripts.sh
@@ -6,6 +6,6 @@
 
 # FIX: Input here should be a set of commands, more precisely, the ones on this specific machine.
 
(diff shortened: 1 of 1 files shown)


angelhof:
That is a good catch! That was exactly the original intention. @dkarnikis: have we been calling `shortest-scripts.sh` with 1G? If so, we might need to change that in evaluation running scripts.

dkarnikis:
@angelhof shortest-scripts are already set to accept https://github.com/binpash/pash/blob/61e5488f2fed07b96d7e2263081fb48db12c67da/evaluation/benchmarks/run.par.sh#L42 as input. I don't think this change affects any of the CI/evaluation script since we ALWAYS export the respective variables

angelhof:
Cool, then @festutz's change only helps for documentation purposes. We can then merge it if things are OK :)

PR #477, opened 2022-02-28
Small documentation fixes

nvasilakis:
This creates a single location where all installation instructions reside, so that they do not fall out of sync. It also fixes several mistakes and adds links to the discord server.


dkarnikis (approved):
Seems good to me!

PR #478, opened 2022-02-28
Draft Python package format

zyrolasting:
I'm inspecting this project's suitability for `pip install pash`. This is just a RFC, because I'm missing key details.

## Changes

* Add `setup.py` to root of repository
* Add blank `__init.py__` to each source directory containing a Python module.

Here, `pip install .` works. The post-conditions need improvement.

## Questions

* Is PyPI the intended host?
* Are all dependencies already known?
* What non-Python files should I include in the package?
* Is [this](https://github.com/binpash/pash/blob/7a26863b97da29bd2ca5e018863eafd3bf97f6f8/compiler/expand.py#L3) intended `import` use if the project is in package form? Or are you looking to use something cleaner like `from pash import ...`?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,13 @@
+import setuptools
+
+setuptools.setup(
+    name='pash',
+    version='0.0.1',    
+    description='Parallelize POSIX shell scripts.',
+    url='https://github.com/binpash/pash',
+    license='MIT',
+    package_dir={'pash': '.'},
+    packages=setuptools.find_packages('.'),
+    install_requires=[],
+    classifiers=[],
+)


tammam1998:
Hi Sage, 

I pushed code to your fork by mistake haha. Not sure how that worked since I shouldn't have write access. I reverted it and added a pull request with some stuff I worked on before that could help you.

@nvasilakis @angelhof can probably answer your questions better but here is my understanding:
* Is PyPI the intended host?
A: My understanding is that this would be a nice to have
* Are all dependencies already known?
A: yes
* What non-Python files should I include in the package?
A:  The main ones at the top of my head are:
- pa.sh is the entry point, 
- script/distro-deps.sh has the Linux packages dependencies
- scripts/setup-pash.sh installs libdash and other debs (I attempt to improve this in scripts/setup-pip.sh)

* Is [this](https://github.com/binpash/pash/blob/7a26863b97da29bd2ca5e018863eafd3bf97f6f8/compiler/expand.py#L3) intended `import` use if the project is in package form? Or are you looking to use something cleaner like `from pash import ...`?
A: I think if we migrate to a clean setup, then it would be better to use `from pash.x.y import z`. Currently, we use packages outside the directory by appending to sys.path which is not very clean. From what I remember, the main problem I faced when I tried to do this is that setup.py needs to be at a directory higher than the project files so we would need to make a significant change to the layout.

zyrolasting:
@tammam1998 Thank you! I'm going to close this PR out after a convo with @nvasilakis, but you'll hear from me again by Friday at the latest.

`pip` is not well suited for system-level deployments. `scripts/install.sh` and company do a good job of handling the challenges, but it's still appears as an extra step that the user might be able to avoid.

I proposed using [FPM](https://fpm.readthedocs.io/en/latest/cli-reference.html#general-options), and you'll see what I'm doing in the next PR. It does everything the current installation scripts do, and spits out `.rpm`, `.deb`, `.pkg` for OSX, etc as output. It accepts both Python packages and system packages as dependencies, and allows me to set a post-install script.

This doesn't make cross-platform deployment _easy_, but much more approachable. For example, a build from source looks like this, assuming `fpm` is installed

```
gh repo clone binpash/pash
cd pash/scripts/package
./build-packages.sh
apt install ./output/pash.deb # Debian
```

The Python paths are still tricky. I'm also a little scared about the part of the code that `mv`s Python modules out of their intended home. I'll try to keep that behavior for compatibility's sake, but I'm okay with moving files around if everyone else is.

PR #479, opened 2022-03-01
Individual aggregation tests

nvasilakis:
Adds individual command tests, for cross-plaform aggregator testing. These are taken from the evaluation scripts.


evaluation/tests/agg/run.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
+
+export PASH_TOP=${PASH_TOP:-$(git rev-parse --show-toplevel --show-superproject-working-tree)}
+# time: print real in seconds, to simplify parsing
+
+bash="bash"
+pash="$PASH_TOP/pa.sh --r_split --dgsh_tee --r_split_batch_size 1000000 --parallel_pipelines --profile_driven"
```
angelhof, line 7: We might want to add `--assert_compiler_success` here, since all of these scripts are pure (I think) and therefore compiler should always succeed. Maybe we also want to add a flag that checks that some parallelization happened (to check that new annotations didn't break parallelization). Currently, all these tests will trivially pass (except if PaSh fails somewhere very early pathologically).

angelhof (commented):
These tests might take a lot of time to run, we need to check before adding them to the CI workflow.

dkarnikis:
> These tests might take a lot of time to run, we need to check before adding them to the CI workflow.

You are correct, there are total 199 tests (20.sh seems to be missing). 
test-123 seems to take a lot of time to complete:
```sh
cat $PASH_TOP/evaluation/tests/input/10M.txt |  xargs -n 2 > out
```
whereas the rest of the suite needs 1m11.373s in total

evaluation/tests/agg/run.sh
nvasilakis: Currently, a relatively small number of tests succeeds (less than 1/3, I believe, from a discussion with @dkarnikis .) The hope is that this will push us to the right direction (add correct cross-platform aggregators), which shouldn't be difficult.

I agree with adding the assertion.

angelhof: Ah nice!!!

dkarnikis: By adding the assertion flag, some tests hang ([test-7](https://github.com/binpash/pash/blob/3a841f6add4bef0f67c10b7e7e6de624316474ec/evaluation/tests/agg/test-7), [test-78](https://github.com/binpash/pash/blob/3a841f6add4bef0f67c10b7e7e6de624316474ec/evaluation/tests/agg/test-78) and possibly more).

In total, 184/197 of the tests pass (without the assertion flag)

janekb04:
There are already some unit tests in /[runtime](https://github.com/binpash/pash/tree/main/runtime)/[agg](https://github.com/binpash/pash/tree/main/runtime/agg)/[cpp](https://github.com/binpash/pash/tree/main/runtime/agg/cpp)/[tests](https://github.com/binpash/pash/tree/main/runtime/agg/cpp/tests) which are ran from `run_tests.sh`. [A quick search](https://github.com/binpash/pash/search?q=run_tests.sh) seems to show though, that currently `run_tests.sh` isn't executed from anywhere anymore. Does it make it obsolete? How do the old unit tests relate to the new tests introduced in this PR?

dkarnikis:
The CI checks are expected to fail since they are running on the old configuration

dkarnikis:
@nvasilakis Should we also remove the `all.sh` file since we already have the testing infrastructure available?

nvasilakis:
> @nvasilakis Should we also remove the `all.sh` file since we already have the testing infrastructure available?

@dkarnikis Sure, this works! I just want to be able to run aggregator tests separately—say, when developing on other platforms such as BSD or MacOS. Is this possible with the current testing infrastructure available?

dkarnikis:
> > @nvasilakis Should we also remove the `all.sh` file since we already have the testing infrastructure available?
> 
> @dkarnikis Sure, this works! I just want to be able to run aggregator tests separately—say, when developing on other platforms such as BSD or MacOS. Is this possible with the current testing infrastructure available?

Yes it is! Just export `PASH_TOP` and then `./run.sh` should do the trick!

dkarnikis (approved):

PR #486, opened 2022-03-08
Fix bug causing graph compilation to fail if termination clean_up_graph flag wasn't set

tammam1998:
If the user doesn't set termination to clean_up_graph, the compiler tries to add a `wait` command to wait for all processes. However, there was a bug in how we added the wait command which causes the compilation to fail. We also didn't report `internal_exec_status` after `wait` which is required at the end of the generated script.

### DIFF
--- a/compiler/ir_to_ast.py
+++ b/compiler/ir_to_ast.py
@@ -115,8 +115,9 @@ def make_ir_epilogue(ephemeral_fids, clean_up_graph, log_file):
         asts.append(com)
     else:
         ## Otherwise we just wait for all processes to die.
-        wait_com = make_command(string_to_argument('wait'))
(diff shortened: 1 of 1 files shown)


angelhof:
That looks OK! Why do we not want to do clean up after termination? Is it because not all graph nodes are on the same machine?

The problem with drain_streams (the other option) is that it doesn't respect `head`'s (and other commands') laziness, meaning that it will run the whole computation even though one one line is actually needed. Therefore, we might want to implement some clean_up for distributed execution correctly. I guess that for now this is fine though.

angelhof (approved):

angelhof:
Feel free to merge whenever you want!

PR #492, opened 2022-03-11
Merge Future

dkarnikis:


dkarnikis:
All the tests fail because:
1. pash setup is different (patched on main)
2. CI configuration is changed (patched on main)

tammam1998:
> All the tests fail because:
> 
> 1. pash setup is different (patched on main)
> 2. CI configuration is changed (patched on main)

Should we merge main into future first then instead of the other way around?

tammam1998 (approved):
Looks good to me

PR #495, opened 2022-03-12
auto split fix

tammam1998:
The hanging issue seems to be resolved if we open all output files at the beginning instead of one by one. From my investigation, I believe that this happens because when tr exists, auto split receives SIGPIPE and exits before opening any of the remaining output files causing other processes to hang. Another solution that seems to work is adding `signal(SIGPIPE, SIG_IGN);` to ignore sigpipe errors and then handle it (open and close remaining fifos) by checking fifo errors.

@angelhof  both solutions seem to work, this pr contains the first one but if you think the second solution is better let me know and I can push it.


angelhof:
This solution looks good to me! @dkarnikis can you check which tests of the new agg tests are supposed to pass and which are supposed to fail so that we didn't introduce any regression? In general, let's disable the failing agg tests until we make them work so that it is clear when merging PRs that we haven't introduced regressions.

Feel free to merge this after we ensure no regressions

dkarnikis:
@tammam1998 When running the `fix-test-191` (this is the same as the issue I created), your fix indeed prevents hanging, however i get the following output:
```sh
tr: when not truncating set1, string2 must be non-empty
tr: when not truncating set1, string2 must be non-empty
```

or
```sh
tr: tr: when not truncating set1, string2 must be non-emptywhen not truncating set1, string2 must be non-empty
```

Whereas bash reports: `tr: when not truncating set1, string2 must be non-empty`

angelhof:
> @tammam1998 When running the `fix-test-191` (this is the same as the issue I created), your fix indeed prevents hanging, however i get the following output:
> 
> ```shell
> tr: when not truncating set1, string2 must be non-empty
> tr: when not truncating set1, string2 must be non-empty
> ```
> 
> or
> 
> ```shell
> tr: tr: when not truncating set1, string2 must be non-emptywhen not truncating set1, string2 must be non-empty
> ```
> 
> Whereas bash reports: `tr: when not truncating set1, string2 must be non-empty`
> 
> The return code for pash is 0 whereas on bash it's 1

This is probably the stderror output (which we never capture or compare with PaSh and it is OK if it is duplicated etc). Furthermore, at the moment I don't think we would return this exit code correctly. We could open an issue for someone to take a look at this and try to figure out a way to return the expected exit code in such cases.

PR #508, opened 2022-03-23
Upload artifact

dkarnikis:
Generate the execution results of posix and correctness tests (including the posix output journal)


dkarnikis:
Merging to main since it doesn't affect pash behavior

PR #518, opened 2022-03-30
HDFS cat support

tammam1998:
Extends the distributed execution mode to handle HDFS cat commands. The PR contributes changes on two levels: 
Compiler: 
- To handle HDFS I introduced a new a`RemoteFileResource` interface and `HDFSResource` implements this interface. The hdfs file resource contains information about a block and all the machines the block exists on.
- HDFSCat DFGNode is added. This node is then removed in the graph optimizing step and replaced by Cat nodes where each cat fid has an HDFSResource that corresponds to a block of the HDFS file.

Distributed Execution:  The worker manager policy now takes into account data availability when choosing a worker.

### DIFF
--- a/compiler/definitions/ir/file_id.py
+++ b/compiler/definitions/ir/file_id.py
@@ -102,6 +102,9 @@ class FileId:
     def has_file_descriptor_resource(self):
         return (isinstance(self.resource, FileDescriptorResource))
 
+    def has_remote_file_resource(self):
+        return isinstance(self.resource, HDFSFileResource)
+
     def is_ephemeral(self):
         return (isinstance(self.resource, EphemeralResource))
 
@@ -118,3 +121,13 @@ class FileId:
 
     def get_ident(self):
         return self.ident
+
+    def is_available_on(self, host):
+        if self.is_ephemeral():
+            return True
+        elif self.has_remote_file_resource():
+            return self.resource.is_available_on(host)
+        else:
+            # Currently any other resource types should
+            # be part of the main shell graph.
+            return False
--- a/compiler/definitions/ir/file_id.py
+++ b/compiler/definitions/ir/file_id.py
@@ -103,7 +103,7 @@ class FileId:
         return (isinstance(self.resource, FileDescriptorResource))
 
     def has_remote_file_resource(self):
-        return isinstance(self.resource, HDFSFileResource)
+        return isinstance(self.resource, RemoteFileResource)
 
     def is_ephemeral(self):
         return (isinstance(self.resource, EphemeralResource))
--- a/compiler/definitions/ir/resource.py
+++ b/compiler/definitions/ir/resource.py
@@ -1,6 +1,7 @@
 from definitions.ir.arg import *
 from util import *
 from ir_utils import *
+import socket
 
 ## TODO: Resources should probably be more elaborate than just a
 ## string and a line range. They could be URLs, and possibly other things.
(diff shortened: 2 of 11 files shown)


compiler/definitions/ir/file_id.py
```diff
@@ -118,3 +121,13 @@ def isNull(self):
...
     def get_ident(self):
         return self.ident
+
+    def is_available_on(self, host):
+        if self.is_ephemeral():
+            return True
+        elif self.has_remote_file_resource():
+            return self.resource.is_available_on(host)
+        else:
+            # Currently any other resource types should
```
angelhof, line 131: And if I understand correctly, the main graph never calls `is_available_on`?

compiler/definitions/ir/resource.py
```diff
@@ -56,3 +57,49 @@ def __eq__(self, other):
...
+        host = addr.rsplit(":", 1)[0]
+        normalized_host = socket.gethostbyaddr(host)[2][0]
+        return normalized_host
+
+class HDFSFileResource(RemoteFileResource):
+    ## The uri is the path of the file.
+    def __init__(self, uri, resource_hosts):
+        """
+        Params:
+            uri: Usually the path to the file
```
angelhof, line 86: Is the path to the file the same in all machines? If not, let's add a TODO here to fix that later.

compiler/definitions/ir/resource.py
```diff
@@ -56,3 +57,49 @@ def __eq__(self, other):
...
+    def __eq__(self, other):
+        if isinstance(other, HDFSFileResource):
+            return self.uri == other.uri
+        return False
+
+    def is_available_on(self, host):
+        return host in self.hosts
+
+    def __repr__(self):
+        return f'hdfs://{str(self.uri)}'
```
angelhof, line 102: I don't think you need to add `str` inside format strings. For example,  `f'{1}` should work and call __repr__ for the value automatically.

compiler/dspash/hdfs_file_data.py
```diff
@@ -32,51 +32,48 @@ def paths(self):
             )
         return filepaths
 
+    def _getinfo(self, filename):
```
angelhof, line 35: It is a bit scary (and unexpected) that when calling the initialization of an object we perform a subprocess call :O. Maybe that can happen in a separate function that is clearly called before initialization, in order to make it clear that here we perform a subprocess call, which could fail, etc...

compiler/definitions/ir/file_id.py
tammam1998: yeah, this function is for evaluating if a fid can be serviced on the provided host. The "main graph" by design runs in the user's shell so we should never call is_available_on on the main graph fids.

The calls for this function are in worker_manager.py

compiler/definitions/ir/resource.py
tammam1998: The path is the same for hdfs; the top directory however could be different. That's why the str for HDFSResource returns "$HDFS_DATANODE_DIR/{self.uri}". I set the HDFS_DATANODE_DIR in the worker's environment.

angelhof (approved):
looks great to me (modulo minor comments)!

compiler/dspash/hdfs_file_data.py
tammam1998: Ok, I could restructure this to be a function call with the object as a return.

compiler/definitions/ir/file_id.py
angelhof: cool!

compiler/definitions/ir/resource.py
angelhof: Great, a comment on that would be nice :)

compiler/dspash/hdfs_file_data.py
angelhof: Sounds good

PR #519, opened 2022-03-30
Badges Table and CI updates

dkarnikis:
* Rename all workflows to `.yaml`
* Smaller workflow names
* Trigger Correctness tests when pushing/merging with main/future (so we can generate the correct badges)
* Update the README to include all the badges for Master(main) and Develop(Future)
* Add new workflow to verify the installation/build process


dkarnikis:
The end goal here is to generate a similar table (we just need to perform some direct pushes on main/future to get the results)

image

(pages is failing since thinkstation is down)

angelhof (approved):

PR #520, opened 2022-03-31
Merge main into future

tammam1998:


dkarnikis (approved):
The changes should not cause any conflict with normal execution since most of them are related to automation pipelines and CI.

angelhof:
I think we can always merge safely from `main` to `future`. No need for approval and stuff. The other way around is what we must be careful with :)

PR #526, opened 2022-04-06
Fix the eager runtime intermediate filenames

a5ob7r:
A previous intermediate filename of the eager runtime is such as `/path/to/tmpdir/prefix_#fifo1`, even though it's a normal file but not a FIFO. This name doesn't make sense, even if it probably has no direct effect on users. So change the filename to indicate that it is a normal file. The fixed filename is such as `/path/to/tmpdir/prefix_1`.

### DIFF
--- a/compiler/pash_runtime.py
+++ b/compiler/pash_runtime.py
@@ -692,7 +692,7 @@ def add_eager(eager_input_id, graph, fileIdGen, intermediateFileIdGen, use_dgsh_
     else:
         ## TODO: Remove the line below if eager creates its intermediate file
         ##       on its own.
-        intermediate_fid = intermediateFileIdGen.next_ephemeral_file_id()
+        intermediate_fid = intermediateFileIdGen.next_temporary_file_id()
 
         eager_exec_path = '{}/{}'.format(config.PASH_TOP, runtime_config['eager_executable_path'])
 
--- a/compiler/definitions/ir/file_id.py
+++ b/compiler/definitions/ir/file_id.py
@@ -39,15 +39,22 @@ class FileId:
         return output
 
     def serialize(self):
-        if(isinstance(self.resource, EphemeralResource)):
+        if(isinstance(self.resource, TemporaryFileResource)):
+            output = self.get_temporary_file_suffix()
+        elif(isinstance(self.resource, EphemeralResource)):
             output = self.get_fifo_suffix()
         else:
             output = "{}".format(self.resource)
         return output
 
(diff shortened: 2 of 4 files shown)


compiler/pash_runtime.py
```diff
@@ -636,9 +636,14 @@ def add_eager(eager_input_id, graph, fileIdGen, intermediateFileIdGen, use_dgsh_
         ## TODO: seperate to better use dgsh-tee params and maybe deprecate eager
         eager_node = dgsh_tee.make_dgsh_tee_node(eager_input_id, new_id)
     else:
-        ## TODO: Remove the line below if eager creates its intermediate file
+        ## TODO: Remove these lines below if eager creates its intermediate file
         ##       on its own.
-        intermediate_fid = intermediateFileIdGen.next_ephemeral_file_id()
+        intermediate_fid = intermediateFileIdGen.next_file_id()
```
angelhof, line 641: Thanks a lot for this change! It might be better if these lines (642 - 646) could be encapsulated behind a `FileId` method (in `definitions/ir/file_id.py`). One clean way I can see of doing it is by having an optional `fifo=True` argument in `next_ephemeral_file_id`. And then we could modify [this function](https://github.com/binpash/pash/blob/main/compiler/definitions/ir/file_id.py#L48) to be named `get_ephemeral_suffix`, which could check whether `fifo` is True or False, and create a different suffix accordingly. Another way, could be to create a different `TemporaryFileResource` and just create that here (and then add all the necessary serialization/to_ast code for it in `file_id.py`.

angelhof (changes requested):
Thanks a lot for this change and sorry for the delay. I have been extremely busy these last few weeks. I suggested a change to keep the code modular and clean (we wouldn't like to introduce a decision about the `eager` file names in this part of the code that only does abstract transformations on the graph). We should ideally move this functionality in `file_id.py`.

Also, would it be possible to target this PR to `future`? Main is only used for bugfixes and new releases (we merge future to main periodically with new releases).

a5ob7r:
Thank you for your review and sorry for the delay my response too.

I think your suggestion makes sense. I'll fix them.

> Also, would it be possible to target this PR to future? Main is only used for bugfixes and new releases (we merge future to main periodically with new releases).

Oh, it's probably right. I change the target to `'future`. Thank you for your pointing.

angelhof:
I run the failing job locally and it passes (it might have to do with the fact that your branch is in your directory).

a5ob7r:
Thank you!

PR #531, opened 2022-04-14
Handle Distributed File System blocks not ending in newline

tammam1998:
Contributes a GRPC server and client and corresponding changes in compiler for handling DFS blocks not ending in newline. This means we use a logical split instead of completely relying on how the HDFS blocks are split.

[design](https://tammammustafa.notion.site/HDFS-newline-ff2aabde3f9e45c0914760c24f164154)


compiler/config.py
```diff
@@ -27,6 +27,8 @@
 
 LOGGING_PREFIX = ""
 
+HDFS_PREFIX = "$HDFS_DATANODE_DIR/"
```
angelhof, line 30: Would we ever need to change this? Would it make sense to set this as a `pash` flag?

angelhof (approved):
Looks good to me!

compiler/config.py
tammam1998: no I don't think we would ever need to change it, so a flag isn't necessary.

PR #532, opened 2022-04-15
Dist functions

tammam1998:
Adds support for functions in distributed execution mode

### DIFF
--- a/compiler/dspash/worker.py
+++ b/compiler/dspash/worker.py
@@ -20,6 +20,7 @@ from annotations import load_annotation_files
 import pash_runtime
 from dspash.socket_utils import send_msg, recv_msg
 from dspash.ir_helper import to_shell_file
+from dspash.utils import create_filename, write_file
 
 # from ... import config
 HOST = socket.gethostbyname(socket.gethostname())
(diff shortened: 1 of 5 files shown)


dkarnikis:
Alright, there was a bug with the POSIX ci when being issued by PRs (works fine when building the website). Should be fixed now :) (you could try re-running all the failed jobs on the other PRs)

compiler/dspash/worker.py
```diff
@@ -41,15 +42,20 @@ def parse_exec_request(request):
...
     script_path = to_shell_file(graph, config.pash_args)
 
     e = os.environ.copy()
     e['PASH_TOP'] = PASH_TOP
-    rc = subprocess.Popen(['bash', script_path], env=e)
+
+    # store functions
+    functions_file = create_filename(dir=config.PASH_TMP_PREFIX, prefix='pashFuncs')
+    write_file(functions_file, functions)
+    cmd = f"source {functions_file}; source {script_path}"
```
angelhof, line 57: Is this the subworker execution script (that executes a part of the graph)? Is there anyway we need to export all functions too?

compiler/pash_runtime.sh
```diff
@@ -153,6 +153,13 @@ else
         exit 1
     fi
 
+    # store functions for distributed execution
+    if [ "$distributed_exec" -eq 1 ]; then
+        declared_functions="${PASH_TMP_PREFIX}/pash_$RANDOM$RANDOM$RANDOM"
+        declare -f > $declared_functions
```
angelhof, line 159: Is this file just sourced by the worker that executes the graph?

angelhof (approved):
Looks good to me!

compiler/pash_runtime.sh
tammam1998: yeah, this file gets stored and the worker manager sends a copy for each worker with the subgraph.

compiler/dspash/worker.py
tammam1998: I don't think we need to re-export because declare -f outputs `declare -fx` statements for exported functions. I tried it and it seems to have the same effect as exporting the functions.

PR #537, opened 2022-04-20
Distribution bug fixes

tammam1998:


compiler/pash_runtime.py
```diff
@@ -116,7 +117,7 @@ def compile_optimize_output_script(ir_filename, compiled_script_file, args, comp
     ## which should be translated to a parallel script.
     if(isinstance(optimized_ast_or_ir, IR)):
         if args.distributed_exec:
-            ir_filename = os.path.join(config.PASH_TMP_PREFIX, 'distributed_exec_ir.pkl')
+            ir_filename = os.path.join(config.PASH_TMP_PREFIX, f'{str(uuid4())[:8]}_distributed_exec_ir.pkl')
```
angelhof, line 120: I think you should use this `https://github.com/binpash/pash/blob/distribution-bug-fixes/compiler/util.py#L42` instead of manually creating random file names.

runtime/dgsh_tee.sh
```diff
@@ -6,12 +6,15 @@ args=${@:3}
 # Set a default DISH_TOP in this directory if it doesn't exist
 PASH_TOP=${PASH_TOP:-$(git rev-parse --show-toplevel)}
 
-cleanup()
-{
-    kill -SIGTERM $dgsh_tee_pid > /dev/null 2>&1
-}
-trap cleanup EXIT
+# TODO: Doable check if this is still needed. Turned off for distributed exection.
```
angelhof, line 9: Could you also add an issue (describing what exactly is happening and why this was disabled)?

angelhof (approved):
Feel free to merge after you do the proposed changes

compiler/pash_runtime.py
tammam1998: The reason I'm not using it is that when debugging (using -d) I want to keep the file but ptempfile always gets deleted when the pash finishes.

angelhof: I thought that when `-d` is used PASH_TMP_PREFIX is not deleted (https://github.com/binpash/pash/blob/main/pa.sh#L79), which would by the way affect your solution too.

Furthermore, ptempfile uses tempfile.mkstemp which is not deleted automatically and requires manual cleanup as far as I know.

runtime/dgsh_tee.sh
tammam1998: Added #539

compiler/pash_runtime.py
tammam1998: Double check and you are right, changed it to use ptempfile.

PR #541, opened 2022-04-26
Add graphviz in setup_pash

angelhof:


angelhof:
Does anyone suspect why CI is not running at all? Here and at #542 ( @dkarnikis if you have any idea that would be helpful)

PR #550, opened 2022-05-10
Fix field splittings by missing double quotations

a5ob7r:
Expansions with unintentional field splitting by shell maybe cause bugs, so we usually wrap them with double quotes to prevent the bugs.

In practice, PaSH has an inconsistent argument handling bug because the current implementation wraps the potential parameter expansion with double quotes at one place but doesn't at another one. For example, PaSH raises a parsing error when running the command below.

```sh
$ pa.sh -c 'cat foo bar | tr -dc abc'
...
pash_runtime_daemon.py: error: argument -d/--debug: invalid int value: 'c'
...
```

To fix this error and prevent similar potential bugs, wrap expansions with double quotes and test using shellcheck to check missing quotations automatically.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -74,3 +74,20 @@ jobs:
           run: |
             # check if everything executed without errors
             cd scripts/workflow && bash exit_code.sh
+
+    shellcheck:
+      runs-on: ubuntu-latest
+      if: github.event.pull_request.draft == false
+      steps:
+        - uses: actions/checkout@v3
+
+        - uses: ludeeus/action-shellcheck@master
+          env:
+            # Only check "Double quote to prevent globbing and word splitting."
+            # now, but we should check others in the future.
+            # https://github.com/koalaman/shellcheck/wiki/Sc2086
+            SHELLCHECK_OPTS: -i SC2086
+          with:
+            ignore_paths:
+              compiler/parser/libdash
+              compiler/tests
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -83,11 +83,22 @@ jobs:
 
         - uses: ludeeus/action-shellcheck@master
           env:
-            # Only check "Double quote to prevent globbing and word splitting."
-            # now, but we should check others in the future.
-            # https://github.com/koalaman/shellcheck/wiki/Sc2086
-            SHELLCHECK_OPTS: -i SC2086
+            # Only check some field splitting problems now, but we should check
+            # others in the future.
+            SHELLCHECK_OPTS:
+              -i SC2046
+              -i SC2048
+              -i SC2053
+              -i SC2068
+              -i SC2086
+              -i SC2206
+              -i SC2254
           with:
             ignore_paths:
+              annotations
               compiler/parser/libdash
               compiler/tests
+              evaluation
+              python_pkgs
+              runtime/agg/cpp/tests
+              scripts
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -97,6 +97,7 @@ jobs:
             ignore_paths:
               annotations
               compiler/parser/libdash
+              compiler/test_evaluation_scripts.sh
               compiler/tests
               evaluation
               python_pkgs
--- a/compiler/pash_runtime.sh
+++ b/compiler/pash_runtime.sh
@@ -140,10 +140,10 @@ else
         fi
         
         # Get assigned process id
-        response_args=($daemon_response)
+        response_args=("$daemon_response")
         process_id=${response_args[1]}
     else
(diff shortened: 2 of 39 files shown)


angelhof:
Thanks a lot for this PR and for finding this bug! The changes look good to me but it would be good to get reviews from the rest of the core team since the changes span multiple files :)

.github/workflows/tests.yaml
```diff
@@ -74,3 +74,20 @@ jobs:
...
+        - uses: actions/checkout@v3
+
+        - uses: ludeeus/action-shellcheck@master
+          env:
+            # Only check "Double quote to prevent globbing and word splitting."
+            # now, but we should check others in the future.
+            # https://github.com/koalaman/shellcheck/wiki/Sc2086
+            SHELLCHECK_OPTS: -i SC2086
+          with:
+            ignore_paths:
```
angelhof, line 91: We probably also need to ignore everything in `evaluation` and all other auxiliary directories.

a5ob7r: OK. I'll add the path to `ignore_paths` after merging https://github.com/a5ob7r/pash/pull/1. In addition, we should add the `scripts` dir to it too?

angelhof: I looked at the ones that fail, and I would suggest adding all of them (`annotations`, `evaluation`, `runtime`, `scripts`). 

We could do another PR on top of this branch of yours so that we check that shellcheck CI passes before we actually merge it here (so that we don't introduce a failing CI workflow).

runtime/split_pipe.sh
```diff
@@ -1,11 +1,11 @@
-$BATCH_SIZE=$1
```
mgree, line 1: Whoa... how did _this_ ever work?

mgree (approved):
Looks good to me. It might be good to add some valid tests that both do and _don't_ quote, to make sure quoting doesn't become load bearing.

runtime/split_pipe.sh
angelhof: Unclear if this is used :')

nvasilakis (approved):
This looks great, thank you!

compiler/pash_runtime.sh
```diff
@@ -140,10 +140,10 @@ else
         fi
         
         # Get assigned process id
-        response_args=($daemon_response)
+        response_args=("$daemon_response")
```
angelhof, line 143: This is a mistake. We want this to be a bash array and so we need to preserve the spaces. @a5ob7r could you revert this? This is the reason why the CI fails.

angelhof (changes requested):
The changes broke pash when we wanted to create a bash array with multiple entries. Could you revert the change that I commented on?

compiler/test_evaluation_scripts.sh
```diff
@@ -173,20 +173,19 @@ execute_tests() {
...
                 echo "|-- Executing with pash --width ${n_in} ${conf}..."
-                export pash_time="${test_results_dir}/${microbenchmark}_${n_in}_distr_$(echo ${conf} | tr -d ' ').time"
+                export pash_time="${test_results_dir}/${microbenchmark}_${n_in}_distr_$(echo "$conf" | tr -d ' ').time"
                 export pash_output="${intermediary_dir}/${microbenchmark}_${n_in}_pash_output"
                 export script_conf=${microbenchmark}_${n_in}
                 echo '' > "${pash_time}"
                 # do we need to write the PaSh output ?
-                cat $stdin_redir |
-                    execute_pash_and_check_diff -d $PASH_LOG $assert_correctness ${conf} --width "${n_in}" --output_time $script_to_execute                 
+                execute_pash_and_check_diff -d "$PASH_LOG" "$assert_correctness" "$conf" --width "$n_in" --output_time "$script_to_execute" <"$stdin_redir"
```
angelhof, line 188: The current problem is somewhere here, though I am not sure where and why...

angelhof: Well, I think I found the problem. We need to remove the quotes from `$assert_correctness` and `$conf` because these contain spaces on purpose.

compiler/test_evaluation_scripts.sh
```diff
@@ -150,18 +150,18 @@ execute_tests() {
         input_file="${prefix}_test.in"
 
         if [ -f "$env_file" ]; then
-            . $env_file
-            vars_to_export=$(cut -d= -f1 $env_file)
+            . "$env_file"
+            vars_to_export=$(cut -d= -f1 "$env_file")
             if [ ! -z "$vars_to_export" ]; then
-                export $vars_to_export
+                export "$vars_to_export"
```
angelhof, line 156: This one needs to be reverted too, since it also contains spaces on purpose.

a5ob7r:
Or we should revert the change of `compiler/test_evaluation_scripts.sh` and add it to `ignore_paths` to skip checking it? Come to think of it, this file is for tests, not for runtime.

angelhof (approved):

PR #558, opened 2022-05-20
Eager fixes for distributed execution

tammam1998:
- Add eager after remote_read commands 
- Fix an issue causing conflicting node ids on the worker. DFGNode.next_id starts from 0 in the worker which causes us to override some nodes when adding eagers. A better long-term solution is to make node id generation local to the graph.


compiler/dspash/ir_helper.py
```diff
@@ -261,6 +265,9 @@ def assign_workers_to_subgraphs(subgraphs:List[IR], file_id_gen: FileIdGen, inpu
                     # sometimes a command can have both a file resource and an ephemeral resources (example: spell oneliner)
                     continue
 
+    # for worker, graph in worker_subgraph_pairs:
```
angelhof, line 268: I would delete this

compiler/pash_runtime.py
```diff
@@ -721,14 +722,20 @@ def add_eager_nodes(graph, use_dgsh_tee):
     intermediateFileIdGen = FileIdGen(0, runtime_config['eager_intermediate_prefix'])
 
     ## Get the next nodes
-    workset = [node for source_node_id in source_node_ids for node in graph.get_next_nodes(source_node_id)]
```
angelhof, line 724: I would be tempted to think that there was a reason why this was here. Maybe we don't want to start from all `source_node_ids`? Is there a particular reason you deleted this? The reason why I am asking is because  regressions here cannot be caught by any test (the compiler won't fail most likely but maybe it won't optimize that much).

compiler/pash_runtime.py
```diff
@@ -721,14 +722,20 @@ def add_eager_nodes(graph, use_dgsh_tee):
...
     while (len(workset) > 0):
         curr_id = workset.pop(0)
         curr = graph.get_node(curr_id)
+
         if (not curr_id in visited):
             visited.add(curr_id)
             next_node_ids = graph.get_next_nodes(curr_id)
+            
+            # Skip if this is the last node
+            if not next_node_ids:
```
angelhof, line 736: Similarly, is that necessary or an optimization?

PR #560, opened 2022-05-25
added limit to the number of parallel pipelines

tammam1998:


compiler/config.py
```diff
@@ -127,6 +127,10 @@ def add_common_arguments(parser):
                         help="Run multiple pipelines in parallel if they are safe to run",
                         action="store_true",
                         default=False)
+    parser.add_argument("--parallel_pipelines_limit",
```
angelhof, line 130: Let's rename this and parallel_pipelines to dependency_untangling now that you are at it :)

angelhof (approved):
Let's run the necessary benchmarks once after this change to check that there is no regression.

compiler/config.py
tammam1998: Planning on keeping it like this for now since the benchmarking scripts use parallel_pipelines. We can change it in a different pull request.

PR #563, opened 2022-06-02
Added annotation for rev

tammam1998:
Needed for NLP benchmark sort_words_by_rhyming (3_3.sh)


angelhof:
This PR introduced a single failing aggregator test in the Debian machine. @janekb04 do you have any idea which one could it be? Can we somehow debug this?

PR #566, opened 2022-06-07
Dspash fixes

tammam1998:
Fixes many bugs and reliability problems:
- Support multiple inputs in hdfs cat "hdfs dfs -cat $IN1 $IN2 .."
- Fix bugs with multi sink scripts
- Support HDFS blocks that are not stored in subdir0 locally
- Merge graph on one worker on the first aggregator (Performance improvement)
- Use processes instead of threads in workers (fixed hanging issue)
- Fixes other bugs and edge cases in graph splitting
- Setup and Installation scripts improvements


angelhof:
I would prefer redirecting these changes to a different branch (other than future). Let's have a dspash-future branch that we will work on temporarily, and when everything stabilizes, we can push to the normal future (to avoid stalling on PRs now). We can still make PRs to this dspash-future branch.

PR #568, opened 2022-06-10
Fix bug in parallel pipelines caused by false reporting of fids

tammam1998:
Fids with no input or output nodes were reported as output fids and input fids causing some conflicts in parallel pipelines (specifically in the distributed execution mode since we replace hdfs cat input fids with an input per block).

Note: this is not a complete solution. replacing HDFSCat with per block cat could cause some conflicts to go undetected. Example: (x | x | x | hdfs dfs -put -f - /1.txt; hdfs dfs -cat) where /1.txt existed before.

### DIFF
--- a/compiler/ir.py
+++ b/compiler/ir.py
@@ -583,19 +583,20 @@ class IR:
 
     ## Returns all the file identifiers in the IR.
     def all_fids(self):
-        all_fids = [fid for fid, _from_node, _to_node in self.edges.values()]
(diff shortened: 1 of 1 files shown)


angelhof:
Hmm, so the issue was there there existed fids with no input or output node in the graph, and so they were returned as input/output fids of the graph right? 

I think the solution should include figuring out who leaks these None to None fids.

BTW, I am not sure I understand the missed conflicts that you are mentioning. We could discuss this out of the critical path however.

PR #576, opened 2022-06-18
Add a note on compatibility in README

angelhof:
I added this note so that new users know what to expect with respect to PaSh's compatibility. @mgree, @nvasilakis feel free to edit or approve this and then we can merge it.


mgree:
Might be good to give an indication of the kinds of things that _don't_ work, e.g., variables in redirects, arrays, process substitutions, some fancy bash parameter forms.

Issue #570, opened 2022-06-15
pa.sh does not handle arguments correctly


ko1nksm:
It seems that pash has a bug in its argument handling.

The script to run is `test.sh`.

```sh
#!/usr/bin/env bash

printf '%s\n' "$@"
```

When executed via `pa.sh`, the space-separated words are the individual arguments.

```console
$  ./test.sh "foo bar" baz
foo bar
baz

$ pa.sh -- ./test.sh "foo bar" baz
foo
bar
baz
```

In addition, pathname expansion is performed.

```sh
$ ls
file1  file2  file3  test.sh

$ pa.sh -- ./test.sh foo '*' baz
foo
file1
file2
file3
test.sh
baz
```

angelhof:
Thanks for reporting these, we will take a look! In general, PaSh has multiple issues with quotes and escaping at the moment, which we are aiming to deal with in a principled way.

angelhof:
These issues are fixed by PR #578 !

angelhof:
Closing since the issues were fixed, but feel free to reopen if you find some other issue related to this, thanks!

PR #578, opened 2022-06-18
[WIP] start fixing argument handling

angelhof:
Working to fix issue: #570 

### DIFF
--- a/compiler/ast_to_ir.py
+++ b/compiler/ast_to_ir.py
@@ -915,8 +915,14 @@ def make_call_to_runtime(ir_filename, sequential_script_file_name,
     ## ```
     assignments = [["pash_input_args",
                     [make_quoted_variable("@")]]]
-    input_args_command = make_command([],
-                                      assignments=assignments)
+    # input_args_command = make_command([],
+    #                                   assignments=assignments)
+    arguments = [string_to_argument("source"),
+                 string_to_argument(config.SAVE_ARGS_EXECUTABLE),
+                 [make_quoted_variable("@")]]
+    ## Pass a relevant argument to the planner
+    # arguments += config.pass_common_arguments(config.pash_args)
+    input_args_command = make_command(arguments)
 
     ## Disable parallel pipelines if we are in the last command of the script.
     ## ```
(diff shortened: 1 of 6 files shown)


angelhof:
The goal is to correctly store input arguments separately (and not join them and resplit them), so that we can keep spaces between them. In addition, we want to restore the arguments without expanding them (so that things like `*` are not expanded to the files in the directory).

I added a test file called `test-star.sh` (in `evaluation/tests/interface_tests`) with this commit (https://github.com/binpash/pash/pull/578/commits/788b9523aa87440f0025fed5c3064dfc81426679) that can observe this issue if called with relevant arguments. For example:

```sh
$PASH_TOP/pa.sh test-star.sh  foo '*' baz "hi there Michael"
```

To resolve this, I came up with a partial solution that uses bash arrays to store the arguments. This would work out of the box if our unparser supported bash syntax, but since it does not, we have to do tricks like putting the bashism in a separate script and sourcing it (since the preprocessor can unparse a script that sources) (see https://github.com/binpash/pash/pull/578/files#diff-9117e6346fe0f0114b82710c6f433eccaeb3e07be02b26f00013024b1400af5a). This makes the preprocessed script look like this:

```sh
{ pash_previous_exit_status="${?}" ; }
 { { source /home/konstantinos/University/research/dish/runtime/save_args.sh "${@}" ; }
 { { pash_disable_parallel_pipelines=0 ; }
 { { source /home/konstantinos/University/research/dish/compiler/pash_runtime.sh /tmp/pash_iQnfOVm/tmpaf611ke8 /tmp/pash_iQnfOVm/tmpjqq8fmc4 --graphviz no --graphviz_dir /tmp --r_split_batch_size 1000000 --debug 1 --termination clean_up_graph --speculation no_spec --width 2 ; }
 { { set -- $(source /home/konstantinos/University/research/dish/runtime/restore_args.sh) ; }
 { ( exit "${pash_runtime_final_status}" ) ; } ; } ; } ; } ; }
{ pash_previous_exit_status="${?}" ; }
 { { source /home/konstantinos/University/research/dish/runtime/save_args.sh "${@}" ; }
 { { pash_disable_parallel_pipelines=1 ; }
 { { source /home/konstantinos/University/research/dish/compiler/pash_runtime.sh /tmp/pash_iQnfOVm/tmpqvb00t8a /tmp/pash_iQnfOVm/tmp1tey1ouu --graphviz no --graphviz_dir /tmp --r_split_batch_size 1000000 --debug 1 --termination clean_up_graph --speculation no_spec --width 2 ; }
 { { set -- $(source /home/konstantinos/University/research/dish/runtime/restore_args.sh) ; }
 { ( exit "${pash_runtime_final_status}" ) ; } ; } ; } ; } ; }
```

while in the `pash_wrap_vars.sh`, I just perform the storing/restoring normally:
```sh
## Store
export pash_input_args=( "$@" )

## Restore
set -- "${pash_input_args[@]}"
```

Unfortunately, I can't find a way to restore using a separate script, i.e., `restore_args.sh`. The best I got was to get it to print them, but this unfortunately ends up adding extra quotes and still splitting different arguments, as follows:

```sh
$PASH_TOP/pa.sh test-star.sh  foo '*' baz "hi there Michael"
foo
*
baz
hi there Michael
"foo"
"*"
"baz"
"hi
there
Michael"
```

This shows that the first storing (in preprocessed script) and the first restoring (in `pash_wrap_vars.sh`) work. We just have to get the restoration in the preprocessed script to work too. 

Some ideas are:
- Can we get the parser to produce the correct restoration somehow, i.e., `set -- "${pash_input_args[@]}"`?
- Could we somehow use `eval` and just produce a string that eval would run?
- Could we correctly print the arguments in `restore_args.sh` so that we can correctly unparse them after the command substitution?

angelhof:
@mgree I think this is done, we should review it on Thursday and merge it!

angelhof:
FYI, the solution involves:
- bash arrays to store arguments (instead of joining/splitting arguments ourselves)
- extensions to our custom expansion to properly parse bash arrays from a `declare`d variables file, and extensions to correctly expand arguments

PR #581, opened 2022-06-22
Installation fixes

nvasilakis:
This fixes a problem with `pexpect` and various other packages that would not be copied appropriately during installation, causing failures.


scripts/distro-deps.sh
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+# TODO: this should be ran before cloning---otherwise cloning fails.
+# It should also set up words etc.
```
angelhof, line 4: What does this mean?

angelhof (approved):
I would suggest making a whitespace change in compiler to trigger the CI before merging.

angelhof:
Looks good to me, can merge :)

PR #588, opened 2022-06-27
add --depth=1 flag to git clone command

averykhoo:
we don't need the entire git history to run pash, so pulling only the latest files speeds things up and reduces the network load (especially as the repo's commit history grows)

Signed-off-by: Avery <averykhoo@gmail.com>


angelhof (approved):
Thanks! It looks reasonable to me :) 

Maybe we don't need to modify the README but just the automated installation scripts (whoever wants to download with `--depth 1` can add it when they follow the instructions.

@nvasilakis If everything looks fine to you too you could merge it!

nvasilakis (approved):

PR #589, opened 2022-06-27
Add support for eager-nodes

festutz:

### DIFF
--- a/compiler/definitions/ir/nodes/eager.py
+++ b/compiler/definitions/ir/nodes/eager.py
@@ -1,118 +1,135 @@
+from datatypes_new.AccessKind import AccessKind
+from datatypes_new.BasicDatatypes import Operand
+from datatypes_new.CommandInvocationWithIOVars import CommandInvocationWithIOVars
+
 from definitions.ir.dfg_node import *
 from ir_utils import *
 
 class Eager(DFGNode):
-    def __init__(self, inputs, outputs, com_name, com_category, com_options = [], 
-                 com_redirs = [], com_assignments=[],
-                 intermediate = None):
-        # BEGIN ANNO : hack for intermediate at the end
-        self.intermediate = intermediate
-        # END ANNO
-        super().__init__(inputs, outputs, com_name, com_category,
-                         com_options=com_options, 
+    def __init__(self,
+                 cmd_invocation_with_io_vars,
+                 com_redirs = [], com_assignments=[]
+                 ):
+        # TODO []: default
+        super().__init__(
+                         cmd_invocation_with_io_vars,
+                         # inputs, outputs, com_name, com_category,
+                         #              com_options=com_options,
                          com_redirs=com_redirs, 
                          com_assignments=com_assignments)
 
     # BEGIN ANNO : copied from DFG node for hack for intermediate at the end
-    def to_ast(self, edges, drain_streams):
-        log(f'do we get here?')
-        ## TODO: We might not want to implement this at all actually
-        if (drain_streams):
-            raise NotImplementedError()
-        else:
-            ## Handle special node to ast here
-            # node = self.special_to_ast(edges)
-            # if node is not None:
-            #     return node
-
-            redirs = self._to_ast_aux_get_redirs()
-            assignments = self.com_assignments
-            ## Start filling in the arguments
-            opt_arguments = []
-            # BEGIN ANNO
-            # get_command_invocation_prefix_from_dfg_node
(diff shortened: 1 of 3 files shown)


compiler/definitions/ir/nodes/eager.py
```diff
@@ -1,118 +1,135 @@
...
-            arguments_new = opt_arguments_new + rest_arguments_backup + [self.intermediate.to_ast()]
-            log(f'arguments_new: {format_args(arguments_new)}')
-            # END ANNO
-
-            all_arguments = [com_name_ast] + arguments_new
-            all_redirs = redirs + new_redirs
-
-            node = make_command(all_arguments, redirections=all_redirs, assignments=assignments)
-        return node
+    # def to_ast(self, edges, drain_streams):
```
angelhof, line 22: Similarly to the previous PR, let's delete those lines instead of commenting them out!

compiler/definitions/ir/nodes/eager.py
```diff
@@ -1,118 +1,135 @@
...
+    cmd_inv_with_io_vars = CommandInvocationWithIOVars(
+        cmd_name=eager_name,
+        flag_option_list=[],
+        operand_list=operand_list,
+        implicit_use_of_streaming_input=None,
+        implicit_use_of_streaming_output=None,
+        access_map=access_map)
+    return Eager(cmd_inv_with_io_vars)
+
+    # eager_name = Arg(string_to_argument(eager_exec_path))
```
angelhof, line 127: These too!

compiler/pash_runtime.py
```diff
@@ -686,12 +688,19 @@ def add_eager(eager_input_id, graph, fileIdGen, intermediateFileIdGen, use_dgsh_
...
     if use_dgsh_tee:
+        assert(False)
         ## TODO: seperate to better use dgsh-tee params and maybe deprecate eager
         eager_node = dgsh_tee.make_dgsh_tee_node(eager_input_id, new_id)
     else:
         ## TODO: Remove the line below if eager creates its intermediate file
         ##       on its own.
+        # TODO: find a better solution to make unique numbers, currently: set to max-value + 1
+        max_old_plus_1 = fileIdGen.next + 1
+        intermediateFileIdGen.next = max_old_plus_1
```
angelhof, line 699: Let's move that to a method of FileIdGen that could be called something like "bump_counter/id/next" or something (instead of setting next itself).

compiler/pash_runtime.py
```diff
@@ -686,12 +688,19 @@ def add_eager(eager_input_id, graph, fileIdGen, intermediateFileIdGen, use_dgsh_
     new_id = new_fid.get_ident()
 
     if use_dgsh_tee:
+        assert(False)
```
angelhof, line 691: This needs to be tested and done because it will soon become the default instead of `eager` (it is even easier than `eager` since it is not given an intermediate FileId if I am not mistaken). You can configure pash to add this using the `--dgsh_tee` flag.

compiler/pash_runtime.py
```diff
@@ -219,13 +219,15 @@ def optimize_irs(asts_and_irs, args, compiler_config):
...
             # log(distributed_graph)
-            
+
             # Eagers are added in remote notes when using distributed exec
             if(not args.no_eager and not args.distributed_exec): 
                 eager_distributed_graph = add_eager_nodes(distributed_graph, args.dgsh_tee)
             else:
                 eager_distributed_graph = distributed_graph
 
+            # print("graph", eager_distributed_graph)
```
angelhof, line 229: No need to add this in the code

compiler/ir.py
```diff
@@ -1073,6 +1073,7 @@ def edge_node_consistency(self):
         for edge_id, (_, from_node_id, to_node_id) in self.edges.items():
             if (not from_node_id is None):
                 from_node = self.get_node(from_node_id)
+                # print("from_node", from_node)
```
angelhof, line 1076: We could remove this before merging

angelhof (approved):
Looks good modulo the minor changes that I wrote. No need to fix `dgsh_tee` in this PR, but it should be fixed for integration to be complete. If we don't fix this in this PR, let's add it to `TODO.md`.

compiler/pash_runtime.py
festutz: I added this to the `TODO`-list.

compiler/pash_runtime.py
festutz: Ok, I have refactored it.

PR #591, opened 2022-06-28
Support for round-robin parallelization with new annotations

festutz:
Including unwrap for commutative commands


angelhof (approved):
Looks good!

PR #595, opened 2022-06-29
Modify dgsh-wrapper ...

festutz:
... to not require operands before options but pass options for input and output directly


festutz:
The wrapper for dgsh_tee uses and modifies PASH_TOP. This is why I decided to amend the wrapper-script to just pass all options (which now also contain input and output) instead of by-passing it. 
If safe to by-pass, it should be easy to do so.

PR #599, opened 2022-06-30
Fuse cat and subsequent split

festutz:

### DIFF
--- a/compiler/ir.py
+++ b/compiler/ir.py
@@ -1,28 +1,18 @@
-# BEGIN ANNO
 import sys
 
 from config import get_path_annotation_repo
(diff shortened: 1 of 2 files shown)


compiler/ir.py
```diff
@@ -1222,6 +1232,8 @@ def edge_node_consistency(self):
             for edge_id in node.get_input_list():
                 _, _, to_node_id = self.edges[edge_id]
                 if(not (to_node_id == node_id)):
+                    log("nodes", self.nodes)
```
angelhof, line 1235: Other than that, this is great!

angelhof (approved):
Great, and so short! Nice!

PR #603, opened 2022-06-30
Install annotations lib using `pip`

angelhof:
"Properly" install and import the annotations library (instead of just adding it to the path).


angelhof:
For tests to actually pass on Ubuntu 18, we need to fix the annotation library `> python3.7` dependency. I will merge for now though because the installation seems to be OK like this.

PR #606, opened 2022-07-02
The new pull request for my changes

qishenli:
I add which and words package for Fedora on the distro-deps.sh to let ./pa.sh ./evaluation/intro/hello-world.sh works.
I also add a Github Action file for the compatibility check. I use Github runner for checking ubuntu-18, 20, and 22 and docker. for checking Fedora(There's no Github Runner for Fedora so I have to use the container.)
Please check. Thanks! @nvasilakis


angelhof:
@qishenli @nvasilakis  is this PR completed? And what is it supposed to do? If it is stale, we can just close it.

PR #607, opened 2022-07-08
Fix CI with new annotation library

angelhof:


angelhof:
I am leaving this PR open until we resolve issues with the annotation-library dependency. At the moment it doesn't work.

PR #623, opened 2022-08-31
Integrate New Annotation Library

festutz:
Things to do before we can merge:
- fix tr-test.sh in testing script: when run outside of the script, pash produces the same results but the testing script reports different outputs
- make sure that CI tests work (cannot be checked for all at the moment since some parts of CI need to be fixed)
- fix commit in annotations-repository for the dependency / deploy the annotations library as a python-library


angelhof:
@festutz we should completely remove the `annotations` directory from this branch and move it under the new annotations repo as `old_annotations` for reference. In any case, it shouldn't be in the PaSh directory anymore.

angelhof:
All the tests pass here! I think we should merge first (since this PR has been waiting here for too long) and after that we can make sure to properly install the annotation library using pypi (instead of cloning it and locally building it using pip). See issue #633

PR #627, opened 2022-09-06
Cleanup setup scripts

angelhof:
Clean up the setup scripts by removing unnecessary components and by preserving all output for better debugability.

### DIFF
--- a/compiler/ir.py
+++ b/compiler/ir.py
@@ -1015,4 +1015,3 @@ class IR:
                 #      and not self.get_stdin() is None 
                 #      and not self.get_stdout() is None)))
 
-
--- a/scripts/distro-deps.sh
+++ b/scripts/distro-deps.sh
@@ -64,9 +64,9 @@ case "$distro" in
         fi
         echo "Running preparation apt install:"
         echo "|-- running apt update..."
-        $SUDO apt-get update &> $LOG_DIR/apt_update.log
+        $SUDO apt-get update #&> $LOG_DIR/apt_update.log
         echo "|-- running apt install..."
-        $SUDO apt-get install -y $pkgs &> $LOG_DIR/apt_install.log
+        $SUDO apt-get install -y $pkgs #&> $LOG_DIR/apt_install.log
         ;;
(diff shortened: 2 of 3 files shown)


angelhof:
@nvasilakis @mgree the setup script no longer sends output to log files to improve debugability (especially when installing PaSh in CI and in docker images). I think this is fine in general too. If you think that is too verbose, we could hide output by default and only show it when a `-v` flag is used. However, I think this is unnecessary for now, and we can just proceed with this PR. Let me know what you think.

mgree:
Looks good to me---I think no need for the flag or anything. It might make sense to include some meta commands to group stuff, as in https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines. But definitely not a big deal.

mgree (approved):
lgtm; i can imagine adding

```
[ "$CI" != "true" ] && exec 3>&1 4>&2 >FOO.log 2>&1
```

at the top of scripts to hide output by default (saving stdout/stderr in fd 3/4 if we need them later for explicit, deliberate, always-on output)

angelhof:
> lgtm; i can imagine adding
> 
> ```
> [ "$CI" != "true" ] && exec 3>&1 4>&2 >FOO.log 2>&1
> ```
> 
> at the top of scripts to hide output by default (saving stdout/stderr in fd 3/4 if we need them later for explicit, deliberate, always-on output)

The problem with this is that it hides all output, right? Would it be better if we instead made a function that can wrap all commands that produce heavy output and hides their output if an environment variable is set, for example:

```sh
hide_if_not_verbose()
{
    if [ "$verbose" == "true" ] || [ "$CI" == "true" ]; then
        "$@"
    else
        >>log.txt 2>&1 "$@"
    fi
}
```

which could be called as follows:

```sh
hide_if_not_verbose python3 -m pip install ...
```

and we can have big progress output always on.

mgree:
This seems reasonable to me. Maybe better to only check `verbose` in the command and default `verbose` when `CI` is set?

PR #651, opened 2023-02-28
Allow preprocessor to extract sequence of commands

angelhof:
Use by running the following when on `$PASH_TOP`.

```sh
./compiler/preprocessor/preprocessor spec input.sh partial_order.sh
```

It will save a sequence of commands in `partial_order.sh.`

For now it is very janky since it prints each command in its own line (but this can fail very badly if commands contain newlines).

TODO: We need to fix this to print each command in its own file, and the partial order file should only contain the ids of the commands and their ordering. I need to add a new directory `speculation` in compiler, that will contain relevant code.

### DIFF
--- a/compiler/pash.py
+++ b/compiler/pash.py
@@ -40,8 +40,7 @@ def main():
         sys.exit(return_code)
 
 def preprocess_and_execute_asts(input_script_path, args, input_script_arguments, shell_name):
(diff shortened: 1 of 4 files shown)


angelhof:
@gliargovas this should be helpful for you!

PR #661, opened 2023-03-10
Python dependency

angelhof:


angelhof:
Don't worry about shellcheck failing, it is accidentally tracking the target branch and not the PR branch. I think I fixed it in the worflow here though

PR #662, opened 2023-03-13
Add In-progress Machine Learning Benchmark

Forthoney:
## Goal
Explore the potential use of PaSh in ML pipelines in this proof of concept benchmark.

## Benchmark Outline
Runs Scikit-Learn's Multiclass Logistic Regression on the [Forest Cover Dataset](https://archive.ics.uci.edu/ml/datasets/Covertype).
It breaks down `LogisticRegression.fit()` into several intermediate steps.

## Details
Logistic Regression is a classification model that aims to find the "best" fitting logistic function for the given training data.

### Input and Output
For a given input, the model passes the input into the trained logistic function. The logistic function will then return a value between 0 and 1 representing the probability that the input belongs to the target class.
image
In the example above, the logistic function of the model predicts that if a student studied for 2 hours, they have a 0.25 probability of passing the exam.

For a binary classification problem (e.g. passed the exam vs. failed the exam), the model will assign a class based just on this probability, since probability that the student failed the exam can be trivially calculated using the property that $Pr[\text{input} \in S] = 1 - Pr[\text{input} \in S^c] $.

For a multiclass classification, the model will have a logistic function for each possible class. Once the model has all the outputs of the logistic function, it will process these values so that they sum to 1, since probabilities must always sum to 1. This is done using something like a [Softmax function](https://en.wikipedia.org/wiki/Softmax_function).

### "Fitting"
The next natural question is how the model learns of an appropriate logistic function.
The model measures the "goodness" of its learned function by calculating the **Loss**, which is calculated by comparing the actual label of an input to the predicted input.
The model will then try to minimize the Loss value.
Different models and tasks require different Loss functions and for Logistic Regression, the [Log Loss function](https://developers.google.com/machine-learning/crash-course/logistic-regression/model-training) is used.

Once it calculates the Loss, it then takes the derivative of all of its parameters with respect to the Loss. This step can be more easily interpreted as the following: the loss is a multivariable function with respect to the model's parameters because if we change the model's internal parameters, the Loss will be different.
So, taking the derivative of the Loss with respect to some parameter will tell us what relationship the two have. If the value is positive, it means increasing that parameter's value will increase the loss. If the value is negative, it means increasing the parameter will decrease the loss. 

Thus, by calculating the Loss's derivative with respect to every weight, the model gains insight into which how parameters should be adjusted.
It applies these adjustments according to a learning rate. With the (hopefully) improved parameter value, the model runs again.

The model repeats this process until the max number of iterations is reached or the derivative of the Loss is 0 (more or less). The latter case suggests that the model is at a local minimum of the Loss which, for various mathematical reasons, is also the global minimum under reasonable, real-life assumptions and constraints.

## Issues/Improvement Points
- Currently only works with Sklearn datasets. For example, Pandas Dataframes will not work.
- Speedup is rather insignificant. Consider replacing `joblib.Parallel` (or its Sklearn wrapper class) with PaSh for further optimizations.

### Appendix
[Slides](https://docs.google.com/presentation/d/1pA3q39Ncph3b5SgOkzFfigiGA49no7tE/edit?usp=sharing&ouid=117053670477944769597&rtpof=true&sd=true) used during meeting


Forthoney:
Created updated pull request

PR #665, opened 2023-04-19
Reorganize pash_runtime

angelhof:


angelhof:
This PR significantly cleans up the pash runtime. It is now ready to make it an external library. @mgree we can discuss the preferred interface of such a library at some point. What would be the most useful interface. I am thinking something along the lines of:
```sh
## Wraps a SCRIPT with a PRE and POST script while making sure to hide this wrapping from SCRIPT
##   by passing it the correct error code and propagating its error code outside while also not resetting `set` state
##   while executing the PRE and POST scripts.
wrap PRE_SCRIPT SCRIPT [POST_SCRIPT]
```

If we can think of a couple usecases too we can add a cute README and publicize this library as its own repo.

PR #677, opened 2023-05-25
Move ast_node under shasta directory to prepare librarification

angelhof:
Also implement printing inside ast_node, stop using libdash


angelhof:
After this PR is merged, we can make this a library. 
We probably want to rename a few things to have consistent clean naming before we do that.

PR #680, opened 2023-05-28
Make `evaluation/intro` more user-friendly

nikpag:
This PR makes the `evaluation/intro` first-time experience a little more straightforward by:

- Adding a README that prompts the user to create required input files beforehand, in case they decide to run the evaluation scripts (such as `demo-spell.sh`) separately, as opposed to running `test.sh` directly.
- Fixing a problem with `curl` that caused the `100M.txt` file to be empty when the data was fetched from www.gutenberg.org.


angelhof (approved):
Thanks for the PR :) It looks great, the issue with the POSIX test has to do with the fact that you did the changes in your own fork (our current CI doesn't support that). Since the change is not pervasive I will just merge!

PR #695, opened 2023-10-31
Remove lists as default arguments in functions

### LINKED ISSUE #485: remove empty lists as default arguments in functions
Let me sketch the problem: Using [] as default argument will only be evaluated once for each function. This is why all function calls will use the same list which is usually not intended.
For more details, see e.g. https://nikos7am.com/posts/mutable-default-arguments/

vagos:
Fixes #485 

Replace lists as default arguments with the `var = [] if var is None else var`  pattern.


angelhof:
Thank you for this change @vagos, can you rebase this commit on `future`? Once you do that I will review and merge :)

vagos:
@angelhof No problem :)

angelhof:
Sorry @vagos, this fell through the cracks! If you can (and want), make a clean PR to future (with just your commit) and ping me so that I can see it and merge it immediately!

vagos:
@angelhof No problem :) On it

PR #697, opened 2023-11-19
Naive Fault Tolerance Support

huangworld:
1. Added naive mock of worker crashes through controlled sleep via worker_timeout and worker_timeout_choice flags.
2. Added naive detection of worker crashes in scheduler based on a retry and backoff logic.
3. Added naive fault tolerance logic in scheduler - re-split the original IR and distribute to healthy workers to (re)-execute again.
4. Upreved benchmarks based on data branch (though require some additional modifications to make these updates suitable for distr context (will be updated in a separate PR in dish repo).

### DIFF
--- a/compiler/config.py
+++ b/compiler/config.py
@@ -174,6 +174,12 @@ def add_common_arguments(parser):
     parser.add_argument("--version",
                         action='version',
                         version='%(prog)s {version}'.format(version=__version__))
+    parser.add_argument("--worker_timeout",
+                        help="determines if we will mock a timeout for worker node.",
+                        default="")
+    parser.add_argument("--worker_timeout_choice",
+                        help="determines which worker node will be timed out.",
+                        default="")
     return
 
 
--- a/compiler/dspash/worker_manager.py
+++ b/compiler/dspash/worker_manager.py
@@ -45,21 +45,34 @@ class WorkerConnection:
         #     answer = self.socket.recv(1024)
         return self._running_processes
 
-    def send_graph_exec_request(self, graph, shell_vars, functions, debug=False) -> bool:
+    def send_graph_exec_request(self, graph, shell_vars, functions, debug=False, worker_timeout=0) -> bool:
         request_dict = { 'type': 'Exec-Graph',
                         'graph': graph,
                         'functions': functions,
                         'shell_variables': None, # Doesn't seem needed for now
-                        'debug': None     
+                        'debug': None,
+                        'worker_timeout': worker_timeout    
                     }
         if debug:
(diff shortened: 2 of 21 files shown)


compiler/config.py
```diff
@@ -174,6 +174,12 @@ def add_common_arguments(parser):
     parser.add_argument("--version",
                         action='version',
                         version='%(prog)s {version}'.format(version=__version__))
+    parser.add_argument("--worker_timeout",
```
angelhof, line 177: I think this should be OK for now, but it would be good if these flags are marked as experimental (in their help) and maybe at some point in the future hidden a bit, or put behind a subgroup so that pash users dont consider using them.

evaluation/benchmarks/max-temp/max-temp-preprocess.sh
```diff
@@ -1,12 +1,12 @@
 #!/bin/bash
 
-sed 's;^;http://ndr.md/data/noaa/;' |
+sed 's;^;atlas-group.cs.brown.edu/data/noaa/;' |
```
angelhof, line 3: these changes were in a different PR, why are they visible here?

angelhof: can you rebase your branch to the correct one (or cherrypick your commits only?)

compiler/dspash/worker_manager.py
```diff
@@ -45,21 +45,34 @@ def get_running_processes(self):
         #     answer = self.socket.recv(1024)
         return self._running_processes
 
-    def send_graph_exec_request(self, graph, shell_vars, functions, debug=False) -> bool:
+    def send_graph_exec_request(self, graph, shell_vars, functions, debug=False, worker_timeout=0) -> bool:
```
angelhof, line 48: I am a little bit concerned that the killing code (sending the worker timeout) is so deeply engrained in the main flow of the pash code. That is OK for now, but is there a way to somehow move this code out of pash? For example, by having a separate entity that is the chaos_manager/fault_orchestrator that sends messages to workers to be killed and so on. This entity can be started when pash starts, but it would make it much simpler to separate the fault causing code from the fault recovery code, and also would make it much easier to remove the faults for pash production use. Let's discuss this in the next meeting!!

evaluation/benchmarks/max-temp/max-temp-preprocess.sh
huangworld: Yes, I'll separate the changes corresponding to benchmark updates to its own PR!

huangworld:
This merge into dspash-ft is reverted. In a separate pr: https://github.com/binpash/pash/pull/707, I retargeted it to ft-future.

PR #701, opened 2023-11-21
initialize libdash only on first call

AnirudhNarsipur:
``` libdash.parser.parse ``` takes in an init argument that indicates whether libdash should be initialized. This should be set to `False` after the first call. Not initializing libdash properly will call the following program to hang:
```
import libdash.parser


a = libdash.parser.parse("bad0.sh")
b = list(a)
c = libdash.parser.parse("bad0.sh")
d = list(c)
```
where `bad0.sh` is simply:
```
cat /a.txt
```

While I believe PaSH only calls libdash once it's better to be defensive.


angelhof (approved):
Looks good to me! If I recall correctly we had discussed this in some meeting some time ago right? Feel free to merge from my side!

PR #706, opened 2023-11-28
Pin pash-annotations version

Forthoney:
PR #705 but for the future branch


angelhof:
@Forthoney  Can you just rebase/merge main to future to get its latest commit instead of duplicating the PR to both?

PR #714, opened 2023-12-16
Move all ArgParser variants to single cli.py module

Forthoney:
I personally had great difficulty navigating multiple different files to figure out how to add flags, modify flags, etc. This PR gathers all the CLI argparsers into a single `compiler/cli.py` file.
This also removes the deprecated flags from the argparsers and changes scripts accordingly.

### DIFF
--- a/compiler/cli.py
+++ b/compiler/cli.py
@@ -1,0 +1,266 @@
+import argparse
+import os
+
+
+class BaseParser(argparse.ArgumentParser):
+    """
+    Base class for all Argument Parsers used by PaSh. It has two configurable flags
+    by default: debug and log_file.
+
+    Other flags are available by classes which inherit BaseParser
+    """
+
+    @staticmethod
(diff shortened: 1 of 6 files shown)


Forthoney:
@angelhof It seems that some tests are failing despite the logic not having changed. My guess is that it fails because of the deprecated options throwing an error - is this a reasonable assuption? If so, I can also add them back in.

angelhof (approved):
Wow, how neat! This is a great PR :) Let's just make sure to readd the obsolete flags and mark them as such (maybe in a function/method that adds them and is clearly documented that they do nothing)! The reason is that it is a breaking change to anyone who uses PaSh to remove flags, and we can do it later at a point when we want to switch to a new major version.

Forthoney:
@angelhof Obsolete flags have been readded under add_obsolete_flags(). Once they are actually phased out we can just not call that method

Issue #632, opened 2022-09-08
Improving exception handling for debugging


festutz:
When debugging the annotations framework, it is sometimes a bit annoying that all exceptions are caught but using `-assert_compiler_success` is not possible for all scripts. I suppose debugging for other parts might have similar issues.
Here, I try to identify the underlying root cause and suggest a possible approach to handle exceptions in a better way.

## Current Situation
The flag `--assert_compiler_success` can be used to require that compilation is successful. Success solely requires that no exceptions was caught (at a certain program point) and does not claim anything about parallelization. For the test cases in `script_microbenchmarks` from `compiler/test_evaluation_scripts.py`, some parts of the regions simply cannot be parallelized, e.g. because of side effects and the "compiler"/parallelization transformations do not fail here in the sense of something has actually gone wrong. Nevertheless, running pa.sh with `--assert_compiler_success` fails and does not output anything since having side effects yields an exception that is caught at the same point as other exceptions.
For debugging, it would be good to be able distinguish why the transformation failed: because of "unparallelizability" (I) (like side effects) or other issues like `IndexError` or `ValueError` (II).

## Suggestion
We could define a dedicated class for Exceptions from (I), throw these at the respective places and catch them before catching all other exceptions which would also cover (II). 
It seems to make sense to change the behavior of `--assert_compiler_success` to only report exceptions (II). 

To get reports for (I) when debugging, one could simply uncomment/introduce `traceback.print_exc()`, which is basically what I do currently for all exceptions, or we could add another flag for that as well.

angelhof:
It seems that this is an issue with many people that are trying to extend pash. I don't have time to do this currently but if anyone wants to get involved it shouldn't be a very hard fix. One approach would be to follow Felix's proposal but maybe always catch all errors normally and only when in debug mode letting type II errors be thrown to the user. Happy to give more context w.r.t. that!

YUUU23:
Hello! 

I started working on this issue a bit by:  

- adding on a custom exception class called `UnparallelizableError` in `custom_error.py` 
- changing some of the general `Exceptions(message)` to this custom exception in `ir.py`, `pash_compiler.py` and `ast_to_ir.py`  
- catch `UnparallelizableError` before all other `Exceptions` in the `compile_ir` function from `pash_compiler.py` 

Here is also the pull request for the above: https://github.com/binpash/pash/pull/720

[...]

PR #720, opened 2024-06-08
Issue 632

YUUU23:
https://github.com/binpash/pash/issues/632


compiler/pash_compiler.py
```diff
@@ -92,10 +93,13 @@ def compile_ir(ir_filename, compiled_script_file, args, compiler_config):
     ret = None
     try:
         ret = compile_optimize_output_script(ir_filename, compiled_script_file, args, compiler_config)
+    except UnparallelizableError as e: 
+        log("WARNING: Exception caught because some region(s) are unparallelizable:", e) 
+        log(traceback.format_exc()) # get detailed reports for (I) when debugging, one could simply uncomment/introduce traceback.print_exc() ?
```
angelhof, line 98: For this we might prefer to leave that commented and make sure to pass all the necessary information in the custom error message. The hope would be that a PaSh user does not see exception traces if something is unparellizable, but rather sees an informative message.

compiler/pash_compiler.py
```diff
@@ -92,10 +93,13 @@ def compile_ir(ir_filename, compiled_script_file, args, compiler_config):
...
     try:
         ret = compile_optimize_output_script(ir_filename, compiled_script_file, args, compiler_config)
+    except UnparallelizableError as e: 
+        log("WARNING: Exception caught because some region(s) are unparallelizable:", e) 
+        log(traceback.format_exc()) # get detailed reports for (I) when debugging, one could simply uncomment/introduce traceback.print_exc() ? 
     except Exception as e:
         log("WARNING: Exception caught:", e)
-        # traceback.print_exc()
-
+        log(traceback.format_exc())
```
angelhof, line 101: Since this will now happen less often it might be OK to let the traceback be logged.

compiler/ir.py
```diff
@@ -960,7 +961,7 @@ def introduce_aggregators_for_consec_chunks(self, fileIdGen, in_aggregator_ids,
                 # TODO: turn node into cmd_invocation_with_io_vars since this is the only thing required in this function
                 self.create_generic_aggregator_tree(original_cmd_invocation_with_io_vars, parallelizer, map_in_aggregator_ids, out_aggregator_id, fileIdGen)
             else:
-                raise Exception("aggregator kind not yet implemented")
+                raise UnparallelizableError("aggregator kind not yet implemented")
```
angelhof, line 964: Maybe add more information in this message. For example, which command does this refer to?

compiler/ir.py
```diff
@@ -741,19 +742,19 @@ def apply_parallelization_to_node(self, node_id, parallelizer, fileIdGen, fan_ou
...
+            raise UnparallelizableError("Splitter not yet implemented")
 
     def apply_round_robin_parallelization_to_node(self, node_id, parallelizer, fileIdGen, fan_out,
                                                   r_split_batch_size):
         # TODO: this control flow should move done to aggregators once we implement them;
         #  currently, this cannot be done since splitter etc. would be added...
         aggregator_spec = parallelizer.get_aggregator_spec()
         if aggregator_spec.is_aggregator_spec_adj_lines_merge():
-            raise Exception("adj_lines_merge not yet implemented in PaSh")
+            raise UnparallelizableError("adj_lines_merge not yet implemented in PaSh")
```
angelhof, line 753: These 3 errors are actually general exceptions. Or rather NotImplemented errors, so I would make a new error type and catch them separately.

compiler/ir.py
```diff
@@ -189,9 +190,9 @@ def compile_command_to_DFG(fileIdGen, command, options,
     command_invocation: CommandInvocationInitial = parse_arg_list_to_command_invocation(command, options)
     io_info: InputOutputInfo = get_input_output_info_from_cmd_invocation_util(command_invocation)
     if io_info is None:
-        raise Exception(f"InputOutputInformation for {format_arg_chars(command)} not provided so considered side-effectful.")
+        raise UnparallelizableError(f"InputOutputInformation for {format_arg_chars(command)} not provided so considered side-effectful.")
     if io_info.has_other_outputs():
-        raise Exception(f"Command {format_arg_chars(command)} has outputs other than streaming.")
+        raise UnparallelizableError(f"Command {format_arg_chars(command)} has outputs other than streaming.")
```
angelhof, line 195: Maybe also print what is the other output?

compiler/ast_to_ir.py
```diff
@@ -141,7 +142,7 @@ def combine_pipe(ast_nodes):
         else:
             ## If any part of the pipe is not an IR, the compilation must fail.
             log("Node: {} is not pure".format(ast_nodes))
-            raise Exception('Not pure node in pipe')
+            raise UnparallelizableError('Not pure node in pipe')
```
angelhof, line 145: Again, I would include the node in the message.

compiler/ast_to_ir.py
```diff
@@ -132,7 +133,7 @@ def combine_pipe(ast_nodes):
     else:
         ## If any part of the pipe is not an IR, the compilation must fail.
         log("Node: {} is not pure".format(ast_nodes[0]))
-        raise Exception('Not pure node in pipe')
+        raise UnparallelizableError('Not pure node in pipe')
```
angelhof, line 136: I would include more information here (like the node), even though it is logged above.

angelhof (commented):
Everything other than the comments looks good. I would suggest rebasing your commits on `future` instead of `main` though. This is where we actually merge new changes (and every now and then we merge on `main`.

PR #721, opened 2024-06-11
Iss632 after rebase on future 

YUUU23:
I might've accidentally messed something up while trying to rebase my old branch to the future branch instead of working off the main branch (😅) ! I apologize for this inconvenience, but here this backup branch should have all the previous debugging improvement changes before Konstantinos's [comments](https://github.com/binpash/pash/pull/720) and also the changes made from Konstantinos's suggestions.


compiler/ast_to_ir.py
```diff
@@ -159,7 +161,7 @@ def combine_pipe(ast_nodes):
     else:
         ## If any part of the pipe is not an IR, the compilation must fail.
         log("Node: {} is not pure".format(ast_nodes[0]))
-        raise Exception("Not pure node in pipe")
+        raise UnparallelizableError("Node: {} is not a pure node in pipe".format(ast_nodes[0]))
```
YUUU23, line 164: Konstantinos (from old branch): I would include more information here (like the node), even though it is logged above.

compiler/ast_to_ir.py
```diff
@@ -168,7 +170,7 @@ def combine_pipe(ast_nodes):
         else:
             ## If any part of the pipe is not an IR, the compilation must fail.
             log("Node: {} is not pure".format(ast_nodes))
-            raise Exception("Not pure node in pipe")
+            raise UnparallelizableError("This specific node: {} is not a pure node in pipe".format(ast_node))
```
YUUU23, line 173: Konstantinos (from old branch): Again, I would include the node in the message.

compiler/ir.py
```diff
@@ -242,11 +243,11 @@ def compile_command_to_DFG(fileIdGen, command, options, redirections=None):
         command_invocation
     )
     if io_info is None:
-        raise Exception(
+        raise UnparallelizableError(
             f"InputOutputInformation for {format_arg_chars(command)} not provided so considered side-effectful."
         )
     if io_info.has_other_outputs():
-        raise Exception(
+        raise UnparallelizableError(
```
YUUU23, line 250: Konstantinos (from old branch): Maybe also print what is the other output?

compiler/ir.py
```diff
@@ -840,7 +841,7 @@ def apply_parallelization_to_node(
                 node_id, parallelizer, fileIdGen, fan_out
             )
         else:
-            raise Exception("Splitter not yet implemented")
+            raise UnparallelizableError("Splitter not yet implemented for command: {}".format(self.get_node(node_id=node_id).cmd_invocation_with_io_vars.cmd_name))
```
YUUU23, line 844: Konstantinos (from old branch): Similarly, add which command the splitter doesn't exist for.

compiler/ir.py
```diff
@@ -849,11 +850,11 @@ def apply_round_robin_parallelization_to_node(
         #  currently, this cannot be done since splitter etc. would be added...
         aggregator_spec = parallelizer.get_aggregator_spec()
         if aggregator_spec.is_aggregator_spec_adj_lines_merge():
-            raise Exception("adj_lines_merge not yet implemented in PaSh")
```
YUUU23, line 852: Konstantinos (from old branch): These 3 errors are actually general exceptions. Or rather NotImplemented errors, so I would make a new error type and catch them separately.

compiler/ir.py
```diff
@@ -1192,7 +1193,7 @@ def introduce_aggregators_for_consec_chunks(
                     fileIdGen,
                 )
             else:
-                raise Exception("aggregator kind not yet implemented")
```
YUUU23, line 1195: Konstantinos (from old branch): Maybe add more information in this message. For example, which command does this refer to?

compiler/pash_compiler.py
```diff
@@ -92,9 +93,15 @@ def compile_ir(ir_filename, compiled_script_file, args, compiler_config):
...
+    except UnparallelizableError as e: 
+        log("WARNING: Exception caught because some region(s) are unparallelizable:", e) 
+        # log(traceback.format_exc()) # uncomment for exact trace report (PaSh user should see informative messages for unparellizable regions) 
+    except (AdjLineNotImplementedError, NotImplementedError) as e: 
+        log("WARNING: Exception caught because some part is not implemented:", e)
+        log(traceback.format_exc())
     except Exception as e:
         log("WARNING: Exception caught:", e)
-        # traceback.print_exc()
+        log(traceback.format_exc())
```
YUUU23, line 104: Konstantinos (from old branch): Since this will now happen less often it might be OK to let the traceback be logged.

compiler/pash_compiler.py
```diff
@@ -92,9 +93,15 @@ def compile_ir(ir_filename, compiled_script_file, args, compiler_config):
         ret = compile_optimize_output_script(
             ir_filename, compiled_script_file, args, compiler_config
         )
+    except UnparallelizableError as e: 
+        log("WARNING: Exception caught because some region(s) are unparallelizable:", e) 
+        # log(traceback.format_exc()) # uncomment for exact trace report (PaSh user should see informative messages for unparellizable regions)
```
YUUU23, line 98: Konstantinos (from old branch): For this we might prefer to leave that commented and make sure to pass all the necessary information in the custom error message. The hope would be that a PaSh user does not see exception traces if something is unparellizable, but rather sees an informative message.

compiler/ir.py
YUUU23: I think I may need some help with determining what the other outputs to pass it along in the exception. I see that in `io_info.has_other_outputs()`, it is going through a for-loop checking if each entry is a "other output" [has_other_outputs() function (line 182)](https://github.com/binpash/annotations/blob/main/pash_annotations/annotation_generation/datatypes/InputOutputInfo.py); I'm not sure if there's already a function in the class that returns these "other outputs", but if not, would it be reasonable to change this function or add another function like this to return a list of those outputs in the class?

angelhof: I am not sure if there is such a function, but if not, you could certainly make a new one that returns a list of those outputs!

compiler/pash_compiler.py
```diff
@@ -92,9 +93,17 @@ def compile_ir(ir_filename, compiled_script_file, args, compiler_config):
         ret = compile_optimize_output_script(
             ir_filename, compiled_script_file, args, compiler_config
         )
+    except ExpansionError as e: 
+        log("WARNING: Exception caught because some region(s) are not expandable and therefore unparallelizable:", e) 
+    except (ExpansionError, UnparallelizableError) as e:
```
angelhof, line 98: Is `ExpansionError` here duplicated on purpose?

python_pkgs/sh_expand/expand.py
```diff
@@ -0,0 +1,518 @@
+import copy
```
angelhof, line 1: This is a different package (https://github.com/binpash/sh-expand)! and the changes in this file need to happen in a PR there.

angelhof (changes requested):
Everything looks good other than the changes to expand.py. These changes need to happen as a new PR to that repo instead (https://github.com/binpash/sh-expand).

compiler/pash_compiler.py
YUUU23: My bad this should be fixed in the newest commit!

python_pkgs/sh_expand/expand.py
angelhof: > Would you recommend for this:
> 
>     1. keeping the ExpansionError in the custom_error and importing it here, or
> 
>     2. moving the ExpansionError here and importing it to pash_compiler to catch it?

I would probably define ExpansionError in sh-expand, and then import it in pash to catch it. It makes sense to hide all expansion related behavior and details in that package.

> I'm not sure why, but it also seems like I don't have permission to push my branch onto this package repo.

That should be fixed now, could you try again? :)

YUUU23: As you suggested, I imported `ExpansionError` into pash_compiler and defined `ExpansionError` in sh_expand in the original package (with a new [PR](https://github.com/binpash/sh-expand/pull/5))! 

I think currently the tests on github actions are not passing due to `ExpansionError` not being found in the package yet. I've ran the tests locally with the most recent changes, and it seems to work as before so hopefully after the change in the package the issue of the `ExpansionError` not found in the import will not be a problem. 

Thank you!

angelhof: I merged that PR (after bumping the project version because that needs to happen for the package to be properly deployed to PyPI). 

Now the tests seem to be passing, so the only thing you need to do is sign off your commits (follow the instructions here) https://github.com/binpash/pash/pull/721/checks?check_run_id=26107013047 and then we can properly approve and merge!

Great work :)

angelhof (approved):
Once you fix the commit signing and the DCO passes, we can squash and merge :)

YUUU23:
> Once you fix the commit signing and the DCO passes, we can squash and merge :)

I fixed the commit signing! Ready for the final steps for merging! Thank you so much!!

angelhof:
Thank you!

PR #722, opened 2024-06-16
feat: --assert_all_regions_parallelizable flag that has the same func…

YUUU23:
To introduce the `--assert_all_regions_parallelizable` flag (should exit with error when both `UnparallelizableError` and general exception are caught) and changing the original `--assert_compiler_success` flag to exiting with error only when general exceptions are caught (does not exit with error when `UnparallelizableError` are caught), the following were added: 

- `compiler/orchestrator_runtime/pash_init_setup.sh`, `compiler/cli.py` : set-up and add information regarding `--assert_all_regions_parallelizable` flag for pash to recognize 
- `compiler/custom_error.py`: introduce `NotAllRegionParallelizableError` to be raised when `UnparallelizableError` are caught in the `compile.ir` function in `pash/compiler/pash_compiler.py` ; 
    - the `NotAllRegionParallelizableError` will be caught in `compiler/pash_compilation_server.py` when the `compile.ir` function is called; if this error is caught, the new variable `all_region_parallelizable` in `compiler/pash_compilation_server.py` will be marked false 
    - to differentiate all regions not parallelizable from general error, a `error_response` with a different error message is saved in the compiler server 
    - don't increase the running procs if `assert_compiler_success` flag is on and `compile_success` is false but `all_region_parallelizable` is true (some general exceptions caught) ; or `assert_all_regions_parallelizable` flag on and `compile_success` is false because of both general exceptions or unparallelizable region 
  
- `compiler/orchestrator_runtime/pash_prepare_call_compiler.sh`: marks variable `pash_all_region_parallelizable` if `error_response` contains message regarding regions not being parallelizable. We should exit on error when, 
   - `assert_all_regions_parallelizable` flag on and `runtime_return_code != 0` : general exception 
   - `assert_all_regions_parallelizable` flag on and `pash_all_region_parallelizable != 0` : some region is not parallelizable 
   - `assert_compiler_success_flag` flag on and `runtime_return_code != 0` and `pash_all_region_parallelizable == 0` : general exception (and not error because some regions were not parallelizable) 

- `evaluation/tests/test_evaluation_scripts.sh` : when executing tests, change `--assert_compiler_success` to `--assert_all_regions_parallelizable`, since what assert_compiler_success originally does is now carried out by `--assert_all_regions_parallelizable`

Tested the above for expected behavior by running some intro scripts in the directory, and it does seem like we are exiting with error for `--assert_compiler_success` only when general exceptions are caught and not the unparallelizable ones. We are also exiting on error for both reasons when `--assert_all_regions_parallelizable` flag is on. When both flags are used at the same time, we follow the behavior of `--assert_all_regions_parallelizable`. PaSh testing scripts locally also seems to pass all tests.

### DIFF
--- a/compiler/cli.py
+++ b/compiler/cli.py
@@ -57,7 +57,12 @@ class BaseParser(argparse.ArgumentParser):
         )
         self.add_argument(
             "--assert_compiler_success",
-            help="assert that the compiler succeeded (used to make tests more robust)",
+            help="assert that the compiler succeeded with no general error occuring",
+            action="store_true",
+        )
+        self.add_argument(
+            "--assert_all_regions_parallelizable",
+            help="assert that the compiler succeeded with all regions being parallelizable and no general error occuring (used to make tests more robust)",
             action="store_true",
         )
         self.add_argument(
--- a/compiler/cli.py
+++ b/compiler/cli.py
@@ -62,7 +62,7 @@ class BaseParser(argparse.ArgumentParser):
         )
         self.add_argument(
             "--assert_all_regions_parallelizable",
-            help="assert that the compiler succeeded with all regions being parallelizable and no general error occuring (used to make tests more robust)",
+            help="assert that the compiler succeeded with all regions being parallelizable and no general error occuring (used to make tests more robust); more strict than --assert_compiler_success flag",
             action="store_true",
         )
         self.add_argument(
--- a/compiler/pash_compilation_server.py
+++ b/compiler/pash_compilation_server.py
@@ -14,6 +14,7 @@ from dspash.worker_manager import WorkersManager
 import server_util
 
 from cli import BaseParser
+from custom_error import * 
 
 ##
 ## A Daemon (not with the strict Unix sense)
@@ -252,6 +253,7 @@ class Scheduler:
         process_id = self.get_next_id()
         run_parallel = False
         compile_success = False
+        all_region_parallelizable = True 
 
         variable_reading_start_time = datetime.now()
         # Read any shell variables files if present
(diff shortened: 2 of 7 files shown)


yizhengx:
LGTM :-)

compiler/cli.py
```diff
@@ -57,7 +57,12 @@ def add_pash_args(self):
         )
         self.add_argument(
             "--assert_compiler_success",
-            help="assert that the compiler succeeded (used to make tests more robust)",
+            help="assert that the compiler succeeded with no general error occuring",
+            action="store_true",
+        )
+        self.add_argument(
+            "--assert_all_regions_parallelizable",
+            help="assert that the compiler succeeded with all regions being parallelizable and no general error occuring (used to make tests more robust)",
```
angelhof, line 65: Maybe we can add a comment that this is strictly stricter than `--assert_compiler_success`

compiler/pash_compilation_server.py
```diff
@@ -321,19 +329,28 @@ def compile_and_add(self, compiled_script_file, var_file, input_ir_file):
...
             self.wait_until_limit(config.pash_args.parallel_pipelines_limit)
-
+        
         if compile_success:
             response = server_util.success_response(
                 f"{process_id} {compiled_script_file} {var_file} {input_ir_file}"
             )
+        elif not all_region_parallelizable: 
+            # send specified message to say not all regions are parallelizable instead of general exception caught
+            response = server_util.error_response(f"{process_id} not all regions are parallelizable; failed to compile")
```
angelhof, line 339: The response can be more specific, and slightly reworded. At this point we know precisely which region is not parallelizable, and we can respond with that (instead of a more generic "not all regions"). Also, the `all_region_parallelizable` flag could be renamed to `current_region_parallelizable` or `parallelizability_success` since it refers to a specific region and not all regions.

angelhof (approved):
Looks great, if you can do the small changes that I suggest we can merge :)

compiler/pash_compilation_server.py
YUUU23: would it be okay to response with "current region" along with the `process_id`? I'm not entirely sure how to refer to a region 
Thank you for the reviews!

angelhof:
Thank you for this fix!!

PR #729, opened 2024-11-12
Add passing bash version to read_vars_file

BolunThompson:
Use the bash version to determine how to parse the vars file as discussed in binpash/sh-expand#7. Should be merged at the same time as those changes (and will fail any tests until then).

### DIFF
--- a/compiler/config.py
+++ b/compiler/config.py
@@ -98,6 +98,9 @@ def load_config(config_file_path=""):
     with open(config_file_path) as config_file:
         pash_config = json.load(config_file)
 
+    # set configuration variables gotten dynamically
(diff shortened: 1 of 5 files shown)


angelhof:
Thanks for this change! We can make it ready for review once the sh-expand library is deployed! Could you change the bash version to be read only once here (https://github.com/binpash/pash/blob/future/pa.sh)? This script runs in bash so it should be able to determine the version and save it in a variable. This is better than hiding the call to bash deep in the compiler.

angelhof:
Thanks @BolunThompson ! Could you sign the commits for the DCO to pass? Then I will merge the PR :)

PR #734, opened 2024-11-18
Set IFS to default before use in compiler.

### LINKED ISSUE #718: Setting IFS crashes PaSh
It seems like someone made a report already here https://github.com/binpash/pash/issues/386 but indeed setting IFS does seem to incorrectly crash PaSh. Here are the steps to reproduce the issue with the following script:

```
IFS=/
echo "hi there"
```

expected behavior when running `$PASH_TOP/pa.sh test.sh`: `hi there`

actual behavior: 

```
/pash/compiler/orchestrator_runtime/pash_prepare_call_compiler.sh: line 51: response_args[1]: unbound variable
```
and then Pash just hangs and must be terminated with a signal.

BolunThompson:
Resolves #718. In a few places, the default value of IFS is assumed, leading to pa.sh crashing when it’s reset. This saves the old IFS and restores it after the word splitting.

### DIFF
--- a/evaluation/tests/interface_tests/test-IFS.sh
+++ b/evaluation/tests/interface_tests/test-IFS.sh
@@ -1,0 +1,5 @@
+IFS=/
+curr_dir=/home/bolun/programming/pash
+for name in $curr_dir; do
+  echo "$name"
+done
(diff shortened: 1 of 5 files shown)


evaluation/tests/interface_tests/test-IFS.sh
```diff
@@ -0,0 +1,5 @@
+IFS=/
+curr_dir=/home/bolun/programming/pash
```
angelhof, line 2: We could "deanonymize" this by making this string to be `/test1/test2/test3/test4`?

angelhof (approved):
Thanks a lot! Very neat! Happy to merge once you deanonymize the test string (see comment) :)

PR #737, opened 2024-12-15
Merge changes in to support parsing bash scripts

BolunThompson:
Code written by @sethsabar. The tests pass with the changes from binpash/shasta#5 and binpash/libbash#1 (CI will fail until those are merged in).


angelhof:
I merged the depending PRs, but I suspect we also need to create a PyPI release of libbash correct?

BolunThompson:
You’re correct — thanks for catching that.

angelhof:
After merging libbash and pushing the repo to PyPI we should be able to continue working on this :)

evaluation/tests/test_evaluation_scripts.sh
```diff
@@ -179,6 +192,9 @@ execute_tests() {
                 export pash_output="${intermediary_dir}/${microbenchmark}_${n_in}_pash_output"
                 export script_conf=${microbenchmark}_${n_in}
                 echo '' > "${pash_time}"
+                if [ "$test_mode" == "bash" ]; then
```
angelhof, line 195: Do we need this here?

evaluation/tests/test_evaluation_scripts.sh
```diff
@@ -96,12 +100,21 @@ pipeline_microbenchmarks=(
 execute_pash_and_check_diff() {
     TIMEFORMAT="%3R" # %3U %3S"
     if [ "$DEBUG" -eq 1 ]; then
-        { time "$PASH_TOP/pa.sh" $@ ; } 1> "$pash_output" 2> >(tee -a "${pash_time}" >&2) &&
-        diff -s "$seq_output" "$pash_output" | head | tee -a "${pash_time}" >&2
+        if [ "$test_mode" == "bash" ]; then
```
angelhof, line 103: I feel that we can just modify the file to run tests both with --bash and without (modifying the configurations array above).

scripts/run_tests_bash.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -x e
```
angelhof, line 3: We should run this script in CI!

evaluation/tests/interface_tests/run.sh
```diff
@@ -328,7 +339,7 @@ test_IFS()
 }
 
 ## We run all tests composed with && to exit on the first that fails
-if [ "$#" -eq 0 ]; then
+if [ "$#" -eq 0 ] || [ "$test_mode" = "bash" ]; then
```
angelhof, line 342: I would really remove all these control flow checks and make sure that all tests always run both with and without --bash.

angelhof (changes requested):
Great job @BolunThompson and @sethsabar !! The only change that we need to do is to make sure that bash tests run in CI (not by modifying the yml file but rather the test scripts). In my opinion, it would be better to just modify all test files to run both with and without `--bash` for all tests (avoiding the control flow that exists currently), however I don't have a strong opinion on that.

BolunThompson:
The current problem is that pash expands words in bash mode by calling `echo {arg}` for every arg, replacing the argument with the output. While this usually works, it doesn’t split words. The dash expansion code, which is necessary to split args before compilation, also naively splits on IFS, leading to commands like `tr “ “ “ “` being parsed as four quotation marks instead of two string arguments (since every character is a CArgChar).

I’m finishing `bash_expand.py`, which sketches out using a bash server to expand asts before compilation, like the dash code. I’ll send (another) PR to sh-expand with it (hopefully soon!).

angelhof:
> The current problem is that pash expands words in bash mode by calling `echo {arg}` for every arg, replacing the argument with the output. While this usually works, it doesn’t split words. Without a bash argument expander, currently they’re just split naively on IFS, leading to commands like `tr “ “ “ “` being parsed as four quotation marks instead of two string arguments (since every character is a CArgChar).
> 
> I’m finishing `bash_expand.py`, which already sketches out using a bash server to expand asts before compilation, like the dash code. I’ll send (another) PR to sh-expand with it (hopefully soon!).

Ohh, that is a bummer... Happy to discuss the solution if needed!

BolunThompson:
I refactored this to use the new (semi-robust) bash expansion server code in binpash/sh-expand#9. Requires #742 to make the tests work, and binpash/shasta#8 for some minor bug fixes.

The bash tests pass on my machine (the dash tests don’t, but I think that’s a problem with my local environment), although a large fraction of the bash tests are commented out. I didn’t test which of those work vs not with the new changes — I can open an issue once this is merged in.

angelhof:
@BolunThompson I think all the PR dependencies are merged! Let me know once this is ready for review and merging!

BolunThompson:
The tests all pass, so this should be ready for review.

angelhof (approved):
Looks great to me! Could we also add a sentence in README saying that if your script includes bash syntax you can use the beta `--bash` flag to support it? Also adding a few paragraphs in docs/bash.md to talk about the architecture of the solution?

BolunThompson:
I wrote some basic documentation — is this good or should I elaborate on some part?

angelhof:
Is the failing test case an expected bug? Can I help debug this somehow? If this is the only thing left let's try to clean it up and merge :)

BolunThompson:
No, unfortunately — I’ll need to look at this a bit more. I think it could
be an issue only with Ubuntu 20.04? It passes on my local Ubuntu 24.04
system.

angelhof:
Hmmm, I merged the PR that updates Ubuntu to 24 to `future`. Could you make another commit to retrigger the tests on the new future? If that doesn't work we can follow up and see what is the issue.

angelhof:
I merged https://github.com/binpash/annotations/pull/22 so once we bump versions here in theory things should be OK?

angelhof:
- I pushed a new commit to `future` to fix a deprecated github action (tight-loop test failure).
- There also seems to be another syntax warning ("/home/runner/work/pash/pash/compiler/dspash/ir_helper.py:82: SyntaxWarning: invalid escape sequence '\ '")
- The hello bash output seems to be different for pash and bash:
```
Running hello-world-bash.sh...hello-world-bash:23.696
1.286
1c1
< 4176
---
> 5971257
```
It is totally unclear to me why :(

BolunThompson:
I was playing around with it, and it only occurs on Ubuntu 22.04 and before. I think it’s an expansion bug related to different bash versions? It works with “${pattern[*]}”, but not with “${pattern[@]}”. On Feb 7, 2025, at 5:30 PM, Konstantinos Kallas ***@***.***> wrote:﻿

I pushed a new commit to future to fix a deprecated github action (tight-loop test failure).
There also seems to be another syntax warning ("/home/runner/work/pash/pash/compiler/dspash/ir_helper.py:82: SyntaxWarning: invalid escape sequence '\ '")
The hello bash output seems to be different for pash and bash:

Running hello-world-bash.sh...hello-world-bash:23.696
1.286
1c1
< 4176
---
> 5971257

It is totally unclear to me why :(

BolunThompson:
Or actually, that’s not it — I forgot that ‘*’ is an unsafe character so it just isn’t parallelized (which makes it work, of course). I’ll look at this more soon.On Feb 7, 2025, at 7:15 PM, Bolun Thompson ***@***.***> wrote:﻿I was playing around with it, and it only occurs on Ubuntu 22.04 and before. I think it’s an expansion bug related to different bash versions? It works with “${pattern[*]}”, but not with “${pattern[@]}”. On Feb 7, 2025, at 5:30 PM, Konstantinos Kallas ***@***.***> wrote:﻿

I pushed a new commit to future to fix a deprecated github action (tight-loop test failure).
There also seems to be another syntax warning ("/home/runner/work/pash/pash/compiler/dspash/ir_helper.py:82: SyntaxWarning: invalid escape sequence '\ '")
The hello bash output seems to be different for pash and bash:

Running hello-world-bash.sh...hello-world-bash:23.696
1.286
1c1
< 4176
---
> 5971257

It is totally unclear to me why :(

BolunThompson:
This just ended up being an environment issue — I’d fixed the bug locally in binpash/sh-expand#11 and forgot to make a new release.

BolunThompson:
The posix correctness tests fail because of the old `actions/upload-artifact` version, but even after a version increment they fail because the `antikythera.csail.mit.edu` [isn’t online anymore](https://github.com/BolunThompson/pash/actions/runs/13229296576/job/36924203594?pr=5). 

(Separately, this action always succeeds because the exit code isn’t captured, so if/when the server comes back online I could add that).

angelhof:
> The posix correctness tests fail because of the old `actions/upload-artifact` version, but even after a version increment they fail because the `antikythera.csail.mit.edu` [isn’t online anymore](https://github.com/BolunThompson/pash/actions/runs/13229296576/job/36924203594?pr=5).
> 
> (Separately, this action always succeeds because the exit code isn’t captured, so if/when the server comes back online I could add that).

@BolunThompson this machine has been down for a long time, and we used to get a report to assess the results (not all tests pass, but a number of them). I don't think it will get back up soon...

Issue #738, opened 2024-12-22
tr-test.sh fails on Ubuntu 24.04


BolunThompson:
Minimal reproduction:
```bash
#!/bin/bash

FILE=$(mktemp)
printf %s 1 > $FILE

cat $FILE
cat $FILE | tr 1 2
cat $FILE
echo # formatting
```

Bash output: `121`
Pash output: `112`

Plausibly related to #736.

Separately, @angelhof thoughts on adding Ubuntu 24.04 (or ubuntu-latest) to CI? There have been a few bugs (#726, #736) that only affect the newer version, and 20.04 is being depreciated in May.

BolunThompson:
I’ll assign this to myself unless someone else wants to take a stab at this.

BolunThompson:
Bizarrely, the script works when invoked as a command substitution. Example:

```bash
./pa.sh test.sh # prints 112
(./pa.sh tr-test.sh) # prints 112
echo $(./pa.sh test.sh) # prints 121
```

Tested with both dash and bash.

angelhof:
> Bizarrely, the script works as intended when invoked as a command substitution. Example:
> 
> ```shell
> ./pa.sh test.sh # prints 112
> (./pa.sh tr-test.sh) # prints 112
> echo $(./pa.sh test.sh) # prints 121
> ```
> 
> Tested with both dash and bash. Does pash do something differently in the last case?

This is really weird, I don't think anything changes when that happens. Does anything look different when using `-d 1`? What do outputs look like?

> Separately, @angelhof thoughts on adding Ubuntu 24.04 (or ubuntu-latest) to CI? There have been a few bugs (https://github.com/binpash/pash/issues/726, https://github.com/binpash/pash/issues/736) that only affect the newer version, and 20.04 is being depreciated in May.

I think that is a great idea! Could you make a PR?

PR #739, opened 2024-12-29
Fix double exit bug in parallel pipelines

### LINKED ISSUE #736: self.running_procs becomes negative, leading to AssertionError
Possibly low priority since I only found the problem while fixing another bug. Minimal test case:
```sh
echo "abc" > /tmp/data.txt
cat /tmp/data.txt
echo
echo
echo
```

Exception:
```python
Traceback (most recent call last):
  File "/home/bolun/pash/compiler/pash_compilation_server.py", line 535, in <module>
    main()
  File "/home/bolun/pash/compiler/pash_compilation_server.py", line 531, in main
    scheduler.run()
  File "/home/bolun/pash/compiler/pash_compilation_server.py", line 510, in run
    self.parse_and_run_cmd(input_cmd)
  File "/home/bolun/pash/compiler/pash_compilation_server.py", line 444, in parse_and_run_cmd
    response = self.compile_and_add(
               ^^^^^^^^^^^^^^^^^^^^^
  File "/home/bolun/pash/compiler/pash_compilation_server.py", line 289, in compile_and_add
    self.wait_unsafe()
  File "/home/bolun/pash/compiler/pash_compilation_server.py", line 433, in wait_unsafe
    assert self.running_procs == 1
           ^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

Checking if self.running_procs already equals zero before decrementing it in `remove_process` does solve the immediate issue, but ideally the number should already match up.

BolunThompson:
Fixes #736, #738. This wasn’t a problem on Ubuntu 20.04 because the EXIT handler wasn’t called (for no obvious reason), so it only exited once.

### DIFF
--- a/compiler/pash_runtime.sh
+++ b/compiler/pash_runtime.sh
@@ -144,7 +144,6 @@ else
             trap inform_daemon_exit SIGTERM SIGINT EXIT
             export SCRIPT_TO_EXECUTE="$pash_script_to_execute"
             source "$RUNTIME_DIR/pash_restore_state_and_execute.sh"
-            inform_daemon_exit
         }
         # Should we redirect errors aswell?
         # TODO: capturing the return state here isn't completely correct. 
--- a/compiler/pash_runtime.sh
+++ b/compiler/pash_runtime.sh
@@ -141,6 +141,8 @@ else
         pash_redir_output echo "$$: (5) BaSh script exited with ec: $pash_runtime_final_status"
     else 
         function run_parallel() {
+            # ideally we'd call inform_daemon_exit with the EXIT handler,
+            # but it isn't called on Ubuntu 20.04 (for no clear reason)
             trap inform_daemon_exit SIGTERM SIGINT
             export SCRIPT_TO_EXECUTE="$pash_script_to_execute"
             source "$RUNTIME_DIR/pash_restore_state_and_execute.sh"
--- a/compiler/pash_runtime.sh
+++ b/compiler/pash_runtime.sh
@@ -141,9 +141,10 @@ else
         pash_redir_output echo "$$: (5) BaSh script exited with ec: $pash_runtime_final_status"
     else 
         function run_parallel() {
-            trap inform_daemon_exit SIGTERM SIGINT EXIT
+            trap inform_daemon_exit SIGTERM SIGINT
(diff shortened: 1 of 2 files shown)


compiler/pash_runtime.sh
```diff
@@ -141,7 +141,7 @@ else
         pash_redir_output echo "$$: (5) BaSh script exited with ec: $pash_runtime_final_status"
     else 
         function run_parallel() {
-            trap inform_daemon_exit SIGTERM SIGINT EXIT
+            trap inform_daemon_exit SIGTERM SIGINT
             export SCRIPT_TO_EXECUTE="$pash_script_to_execute"
             source "$RUNTIME_DIR/pash_restore_state_and_execute.sh"
             inform_daemon_exit
```
angelhof, line 147: Maybe the best solution would be to leave the trap as is, and remove the inform at the end of the call to run_parallel? This would make sure that we only call `inform_daemon_exit` once (I wonder whether the current implementation allows a subtle corner case now where SIGTERM is received while we are in inform). 

Based on your explanation (EXIT not being called on Ubuntu 20) my proposed fix shouldn't work on Ubuntu 20, so if we agree that this would be a better fix but it doesn't work on Ubuntu 20, let's leave a comment in the code to remember that we have it at the end and at the trap because for some reason the trap is not properly triggered, but if it was, that is how we would like to implement it

angelhof (approved):
Could you add read my comment and if you agree add a comment in the code so that we remember that? After this is done (or if my comment doesn't make sense) we will merge :)

compiler/pash_runtime.sh
BolunThompson: That’d be the best solution, but it doesn’t work, unfortunately. State of the repo after the first failed CI run: https://github.com/binpash/pash/pull/739/files/509fcf00f62edd210d969708dfee0d26e4088919

I’ll add a comment mentioning this. I can add a quick check in inform_daemon_exit as for whether exit it’s been called already to avoid the sigterm corner-case, which’d avoid a future bug like this.

angelhof: A comment should be fine :) No need to solve it since it wasn't solved previously either!

PR #740, opened 2024-12-29
Fix typo in compile_asts

BolunThompson:


angelhof:
Hmm, did this manifest as a bug somewhere?

BolunThompson:
It doesn’t, now that I look at it. It’s from an older version where compiled_asts returned the AST to prevent compilation (instead of raising an exception). I removed the code.

angelhof:
Very cool, thanks!

PR #742, opened 2025-01-11
Fix —assert-* checks

BolunThompson:
1. Call inform_daemon_exit on —assert_* exit. Otherwise, --assert_compiler_success hangs while pash waits for an exit message that never comes. The tests don’t currently use this flag, but it will be used for #739.

2. Corectly parse is_parallelizable output. Doesn’t affect the tests, since --assert_all_regions_parallelizable failed anyways on the non-zero error code.


angelhof:
Great job!

PR #744, opened 2025-01-19
fix quoted commands not being parallelizable

### LINKED ISSUE #741: Quoted commands aren’t recognized as being parallelizable.
Example: `"cat" "$0" | "tr" "A-Z" "a-z"` isn’t parallelizable, but `cat $0 | tr A-Z a-z` is. Not a common scenario, but should be a simple fix.

kwakubiney:
Attempt to fix #741


compiler/annotations_utils/util_parsing.py
```diff
@@ -66,8 +66,10 @@ def parse_arg_list_to_command_invocation(
     command, flags_options_operands
 ) -> CommandInvocationInitial:
     cmd_name = format_arg_chars(command)
+    #strip quotes from command where necessary
+    #quoted commands are interpreted as non-parallelizable
+    cmd_name = cmd_name.strip().strip('"').strip("'")
```
angelhof, line 71: I think it is better to do this change by making a new function replacing `format_arg_chars` since that function has access to the actual character codes and not just the string, making any transformation more robust! At this point in the execution, it should certainly be the case that `command` only is expanded and only has quotes (`Q` characters), normal (`C`), and escaped (`E`) characters I think. So we can make new function called `remove_quotes_expanded_arg_chars` that:
1. First checks if the word is indeed expanded, only containing the characters above (to do this we need to traverse the whole argument hierarchically.
2. Then removes the outer layers of quotes (if any).

angelhof (changes requested):
Thanks for the fix and for adding a test!

In addition to the change that I added in the text, it would also be great if you could rebase this to the `future` branch, since we always push new changes to future and then later push them to `main` everytime we do a release!

compiler/annotations_utils/util_parsing.py
kwakubiney: I attempted this in my latest commit but seems to be breaking some tests. Not sure if I am making a obvious error

PR #745, opened 2025-01-25
Add new Ubuntu version to CI tests

BolunThompson:

PR #746, opened 2025-01-25
Upgrade CI to latest OS versions

BolunThompson:
Separate from #745 in case we want to wait and see if it causes any problems.

Separately, the docker containers are a bit out of date (last pushed May 2023), so could whoever has permissions to update those push the new version? (although I’m not sure how important those are to keep up to date).

### DIFF
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -11,7 +11,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - ubuntu-20.04
+          - ubuntu-24.04
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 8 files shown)


angelhof (approved):
Great! What is the order in which this should be merged with #737 #745? Regarding the docker containers, could you make an issue so that we don't forget?

BolunThompson:
This is independent from #737. It could be merged at the same time as #745 (or afterwards) if you think it’d be fine to switch docker to 24.04 simultaneously (although in hindsight we could just choose to push a new docker image later on if pash seems to work with ubuntu 24.04 — it works locally, at least).

angelhof:
I am not sure there is any dependency, but I would like to merge this after #737 so that we make sure that passes first (I don't want to keep it lingering for long :)

PR #750, opened 2025-02-24
Local annotations setup and minor dependency handling

DivikChotani:
- Modified `pa.sh` for compatibility with local annotations
- Added `local-annotations-library-documentation.md` with setup instructions
- Introduced `local-or-pypi.sh` to configure package installation mode for local annotations

Signed-off-by: Divik Chotani divikchotani@g.ucla.edu


local-annotations-library-documentation.md
```diff
@@ -0,0 +1,83 @@
+# PaSh Setup and Execution Guide
```
angelhof, line 1: This file should be moved under `docs` and we should add a pointer sentence to it from the README :)

local-annotations-library-documentation.md
```diff
@@ -0,0 +1,83 @@
...
+- Exporting `PASH_TOP`
+- Running the custom `local-or-pypi.sh` script to configure package installation
+
+---
+
+## **🔹 Step 1: Clone PaSh Repository (If Not Already Cloned)**
+If you haven't already cloned PaSh, do so with:
+
+```sh
+git clone https://github.com/binpash/pash.git ~/pash
```
angelhof, line 16: No need to clone it in `~/pash`, you can just say 
```
git clone https://github.com/binpash/pash.git
cd pash
```

local-annotations-library-documentation.md
```diff
@@ -0,0 +1,83 @@
...
+
+Run:
+
+```sh
+./scripts/setup-pash.sh
+```
+
+
+
+Do not forget to export before using pash: `export PASH_TOP=/home/your-username/pash`
```
angelhof, line 42: this should be changed to the directory that pash is stored, no need to specify it here :)

local-annotations-library-documentation.md
```diff
@@ -0,0 +1,83 @@
...
+To make this persistent across terminal sessions, add it to your shell configuration file:
+
+```sh
+echo 'export PASH_TOP=/home/your-username/pash >> ~/.bashrc
```
angelhof, line 46: There is a missing single quote here

local-annotations-library-documentation.md
```diff
@@ -0,0 +1,83 @@
...
+PaSh provides a script to configure whether dependencies should be installed locally (using a cloned repository) or from PyPI.
+
+To use local dependencies:
+
+```sh
+./local-or-pypi.sh local
+```
+
+This will:
+- Check if `~/annotations` exists.
```
angelhof, line 60: Ideally it should also take an annotation repository as a path (or assume it is a sibling to PaSh but not that it is in the home directory).

local-annotations-library-documentation.md
```diff
@@ -0,0 +1,83 @@
...
+- If not, clone it from GitHub.
+- Update `requirements.txt` to use the local `annotations` repo.
```
angelhof, line 62: Do we save the old backup requirements anywhere? Can't we create a new requirements file maybe?

local-annotations-library-documentation.md
```diff
@@ -0,0 +1,83 @@
...
+
+To use PyPI dependencies:
+
+```sh
+./local-or-pypi.sh pypi
```
angelhof, line 67: Why do we need this? Isn't this what happens by default in the existing installation scripts?

local-annotations-library-documentation.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+
+## **🔹 Step 6: Run PaSh**
+Once everything is set up, test PaSh with:
+
+```sh
+cd ~/pash/evaluation/intro
```
angelhof, line 75: You can also propose to really run the actual tests `https://github.com/binpash/pash/blob/main/evaluation/tests/test_evaluation_scripts.sh`.

local-or-pypi.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+# Define the requirements file path
+REQ_FILE="requirements.txt"
+ANNOTATIONS_DIR="$HOME/annotations"
```
angelhof, line 5: Make this a parameter :)

local-or-pypi.sh
```diff
@@ -0,0 +1,58 @@
...
+
+# Check if an argument is provided
+if [ $# -ne 1 ]; then
+    echo "Usage: $0 <local|pypi>"
```
angelhof, line 9: Do we want the pypi version?

local-or-pypi.sh
angelhof: Or by default the sibling of the pash directory (but not necessarily in the home).

local-or-pypi.sh
```diff
@@ -0,0 +1,58 @@
...
+fi
+
+# Write the correct `requirements.txt`
+if [ "$1" == "pypi" ]; then
+    echo "📦 Writing PyPI requirements to $REQ_FILE..."
+    cat <<EOF > "$REQ_FILE"
+graphviz
+libdash
+pash-annotations==0.2.2
+shasta==0.1.0
```
angelhof, line 39: We don't want to rewrite these here, we better just take them from the requirements file

local-or-pypi.sh
```diff
@@ -0,0 +1,58 @@
...
+sh-expand>=0.1.6
+-e $ANNOTATIONS_PATH
+EOF
+else
+    echo "❌ Invalid argument: $1"
+    echo "Usage: $0 <local|pypi>"
+    exit 1
+fi
+
+echo "✅ Requirements file updated successfully."
```
angelhof, line 57: Maybe this file should instead not change the requirement but rather install the local annotations repo. That makes me wonder whether it should be a part of `setup-pash.sh` and install local annotations when passed a specific flag, like `--local-annotations`.

pa.sh
```diff
@@ -3,7 +3,10 @@
 export PASH_TOP=${PASH_TOP:-${BASH_SOURCE%/*}}
 export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/"
 # point to the local downloaded folders
+export PYTHONPATH=$PASH_TOP/venv/lib/python3.12/site-packages:$PYTHONPATH
```
angelhof, line 6: Is this necessary or can the use of venv be split in a different PR?

pa.sh
```diff
@@ -3,7 +3,10 @@
...
 export PYTHONPATH="${PASH_TOP}/python_pkgs/:${PYTHONPATH}"
+export PYTHONPATH=~/annotations:$PYTHONPATH
```
angelhof, line 8: This is too hardcoded. Maybe the default value is next to the pash directory (sibling of PaSh top).

requirements.txt
```diff
@@ -1,5 +1,5 @@
 graphviz
 libdash
-pash-annotations==0.2.2
 shasta==0.1.0
 sh-expand>=0.1.6
+-e /home/divikchotani.linux/annotations
```
angelhof, line 5: this doesn't look right :)

angelhof (changes requested):
Thanks a lot for the changes Divik :) Here are some comments that you can address before we look at this again. Two more things:
1. You should retarget this PR to go to `future` instead of `main`.
2. You should add a test of this process, this could be next to the other tests that PaSh runs and then add this test to CI too (check how to add CI tests here: `https://github.com/binpash/pash/blob/main/.github/workflows/tight-loop.yaml`)

pa.sh
```diff
@@ -4,6 +4,24 @@ export PASH_TOP=${PASH_TOP:-${BASH_SOURCE%/*}}
 export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/"
 # point to the local downloaded folders
 export PYTHONPATH="${PASH_TOP}/python_pkgs/:${PYTHONPATH}"
+export ANNOTATIONS_PATH=$([ -f "$PASH_TOP/local-annotations-path.txt" ] && tr -d '[:space:]' < "$PASH_TOP/local-annotations-path.txt")
```
angelhof, line 7: I see where you are going with this but it is very scary and a very big dependency for PaSh. I think the most appropriate way would be to modify PaSh's installation (scripts/setup_pash.sh) to copy the annotation repo in the right place. I could see two ways to do this:
1. Either create a small installation script that replaces the annotation repo where pash looks for dependencies (https://github.com/binpash/pash/blob/future/scripts/setup-pash.sh#L37) and call it there (https://github.com/binpash/pash/blob/future/scripts/setup-pash.sh#L37). Then everytime the user modifies the annotation repo they will call this script and replace their local version of annotations. If we do that we also need to create a script that redownloads the latest normal annotations if needed.
2. Properly add a new `--local-annotations-dir` option to PaSh (https://github.com/binpash/pash/blob/future/compiler/orchestrator_runtime/pash_init_setup.sh#L27) which could be used to update PYTHONPATH to point first to where the local annotation library is. I think this is the preferred solution if it works :)

angelhof:
I left some more suggestions for changes. Once these are done, it would also be great to check that this works. That is why adding a test in CI would be great to guarantee we never break this. To do this, we would add a CI file similar to this: https://github.com/binpash/pash/blob/future/.github/workflows/tight-loop.yaml where we download and install pash (similar to this CI file), and then 
1. download annotation library
2. add a new annotation
3. execute a script that uses this annotation and assert that it succeeds

I understand that this is a lot of work, so this could become a different PR. Let's try to fix the current issues now and merge this PR and then we can see how to make this work.

compiler/orchestrator_runtime/pash_init_setup.sh
```diff
@@ -87,11 +88,27 @@ do
...
+        #echo "✅ Using local annotations directory: $ANNOTATIONS_PATH"
+    fi
 done
 
 ## `pash_redir_output` and `pash_redir_all_output` are strictly for logging.
 ##
 ## They do not execute their arguments if there is no debugging.
+if [ "$local_annotations" -eq 1 ]; then
+    # Comment out the PyPI-installed `pash-annotations`
+    sed -i 's/^pash-annotations.*/#&/' "$PASH_TOP/requirements.txt"
```
angelhof, line 106: Why do we need to do that? At this point is anyone reading the requirements.txt?

compiler/orchestrator_runtime/pash_init_setup.sh
```diff
@@ -87,11 +88,27 @@ do
     if [ "--distributed_exec" == "$item" ]; then
         export distributed_exec=1
     fi
+
+    if [ "--local-annotations-dir" == "$item" ]; then
+        shift
+        export ANNOTATIONS_PATH=$(realpath "$2")
```
angelhof, line 94: I found it surprising that this works with $2 here...

docs/local-annotations-library-documentation.md
```diff
@@ -0,0 +1,76 @@
...
+
+To run it with local annotation:
+```sh
+cd pash/evaluation/intro
+time $PASH_TOP/pa.sh $PASH_TOP/evaluation/intro/hello-world.sh --local-annotations
+```
+
+You can also run the full PaSh test suite:
+```sh
+$PASH_TOP/evaluation/tests/test_evaluation_scripts.sh
```
angelhof, line 71: This doesn't run the PaSh test suite with the local annotations, why do you mention it then?

local-annotations-path.txt
```diff
@@ -0,0 +1 @@
+/home/divikchotani.linux/annotations
```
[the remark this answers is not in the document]
angelhof: Or completely removed since it is not used anymore

pa.sh
angelhof: Why do we need the changes in pa.sh now? I think we should delete all of them, I don't see why they are necessary.

docs/local-annotations-library-documentation.md
```diff
@@ -0,0 +1,76 @@
...
+
+```sh
+echo 'export PASH_TOP=/path/to/pash' >> ~/.bashrc
+source ~/.bashrc
+```
+
+## **🔹 Step 3: Run setup-annotations.sh to Configure Package Installation**
+PaSh provides a script to configure dependencies, including an option to install local annotations.
+
+To install dependencies normally:
```
angelhof, line 44: Is this a typo?

setup-local.sh
```diff
@@ -0,0 +1,37 @@
...
+if [ $? -ne 0 ]; then
+    echo "❌ Failed to clone annotations repository."
+    exit 1
+fi
+
+# Get the absolute path
+ANNOTATIONS_PATH=$(realpath "$ANNOTATIONS_DIR")
+echo $ANNOTATIONS_PATH > local-annotations-path.txt
+echo "✅ Annotations repository cloned to: $ANNOTATIONS_PATH"
+grep -qxF "$ANNOTATIONS_PATH" requirements.txt || echo -e "\n-e $ANNOTATIONS_PATH" >> requirements.txt
```
angelhof, line 36: Why do we modify requirements here? Why is this useful?

compiler/cli.py
```diff
@@ -215,6 +215,14 @@ def __init__(self, *args, **kwargs):
             help=" output the preprocessed script",
             action="store_true",
         )
+        self.add_argument(
```
angelhof, line 218: This is only useful for the help message, other than that the parsing of the argument just happens outside in pash_init_setup.sh so I am not sure whether all of these options (like nargs, const) are useful

angelhof (changes requested):
Thanks, please carefully address the changes that I marked here and let me know if you need more context for any of them :)

docs/local-annotations-library-documentation.md
```diff
@@ -0,0 +1,139 @@
...
+To run it with local annotation:
+```sh
+cd pash/evaluation/intro
+time $PASH_TOP/pa.sh $PASH_TOP/evaluation/intro/hello-world.sh --local-annotations-dr /path/to/annotations
+```
+
+For more details, check the test script here:  
+[PaSh Evaluation Tests](https://github.com/binpash/pash/blob/main/evaluation/tests/test_evaluation_scripts.sh)
+
+# How to Add a New Command to PaSh's Annotation System
```
angelhof, line 62: This is very useful!! Can we add these instructions as a PR to the annotations repo and point to them from here?

compiler/pash.py
```diff
@@ -98,7 +98,7 @@ def parse_args():
     if args.command is not None:
         fname = ptempfile()
         with open(fname, "w") as f:
-            f.write(args.command)
+            f.write(args.command + "\n")
```
angelhof, line 101: scary impactful change (why is this necessary?)

compiler/orchestrator_runtime/pash_init_setup.sh
```diff
@@ -87,11 +89,20 @@ do
     if [ "--distributed_exec" == "$item" ]; then
         export distributed_exec=1
     fi
+
+    if [ "--local-annotations-dir" == "$item" ]; then
+        count=$(( count + 1 ))
```
angelhof, line 94: no need to use count, see how other options are parsed! Like in line 73. We first set a variable that the next argument is the option argument, and then we check that argument in the next iteration of the loop.

compiler/orchestrator_runtime/pash_init_setup.sh
```diff
@@ -221,6 +232,7 @@ else
     {
         local daemon_pid=$1
         ## Only wait for daemon if it lives (it might be dead, rip)
+
```
angelhof, line 235: This change seems irrelevant so could be removed. Random modifications in the file make it harder to understand what is meaningful change.

compiler/orchestrator_runtime/pash_prepare_call_compiler.sh
```diff
@@ -41,6 +41,7 @@ fi
 
 if [[ "$daemon_response" == *"OK:"* ]]; then
     pash_runtime_return_code=0
+    at_least_one_optimization=1
```
angelhof, line 44: where is this used and what is it?

compiler/orchestrator_runtime/pash_restore_state_and_execute.sh
```diff
@@ -14,6 +14,7 @@ export pash_current_set_state=$-
 source "$RUNTIME_DIR/pash_set_from_to.sh" "$pash_current_set_state" "$pash_previous_set_status"
 pash_redir_output echo "$$: (3) Reverted to BaSh set state: $-"
 
+
```
angelhof, line 17: please no random whitespace changes :)

scripts/setup-annotations.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+
+# Determine PaSh top directory
+PASH_TOP=$(dirname "$(realpath "$0")")
```
angelhof, line 5: do we want to get the original PASH_TOP value if it is set already? Only if it is not set we want to set it based on this. Also, this setting here assumes we are calling the script from the root of the repo. I would suggest two modes:
1. if we have an extra argument with user provided path just execute there
2. if not, then check if PASH_TOP is set and if not print error message and exit. If it is set, we can install as a sibling

scripts/setup-annotations.sh
```diff
@@ -0,0 +1,33 @@
...
+    ANNOTATIONS_DIR=$(realpath "$ANNOTATIONS_DIR")
+else
+    ANNOTATIONS_DIR="$DEFAULT_ANNOTATIONS_DIR"
+fi
+
+echo "Annotations directory set to: $ANNOTATIONS_DIR"
+
+# Always clone the repository to the specified or default location
+echo "Downloading annotations repository..."
+rm -rf "$ANNOTATIONS_DIR"
```
angelhof, line 22: I would not do that by default, I would either add a flag to remove (like --force or -f) or print an error if the directory already exists.

scripts/ci/test_local_annotations.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+# Define the file path
+FILE="$PASH_TOP/../annotations/pash_annotations/annotation_generation/AnnotationGeneration.py"
+
+# First run the script normally, make sure that the optimization occurred by checking that the FIFOs were made
+$PASH_TOP/pa.sh $PASH_TOP/evaluation/intro/hello-world.sh -d 2 --local-annotations-dir ~/annotations/ --log_file "$PASH_TOP/pash.log"
```
angelhof, line 7: If you test on a script with a single pipeline, you can use this flag to return whether it succeeded or not `--assert_compiler_success`. The flag is also used in other tests so you can find how to use it instead of using grep :)

.github/workflows/annotations.yaml
```diff
@@ -0,0 +1,58 @@
+name: annotations
+
+on:
+  push:
+    branches:
+      - future
+      - main
```
angelhof, line 7: Great, have you tested whether this works on your own fork? If you add it to your own fork and then do a PR on top of it, this should run too, so you should be able to test that it works properly :)

angelhof (changes requested):
Thanks for the changes :) We are almost there, we are getting very close to something that is mergeable! Thanks a lot for the effort!

DivikChotani:
@angelhof Hi Professor Kallas, I believe I have made all the requested changes, and I have tested the CI script from pushing onto my own fork and it works perfectly. I have also cleaned up the code and got rid of a tiny merge conflict that was occurring on my side. Please let me know If I have any other changes that should be made, thanks!

I also made a sibling pr on the annotations repo where the documentation to add commands is now

.github/workflows/annotations.yaml
```diff
@@ -0,0 +1,56 @@
...
+          export PASH_TOP=$PWD
+          export PATH=$PATH:$PASH_TOP
+
+      - name: Run Custom Test Scripts
+        run: |
+          export PASH_TOP=$PWD
+          export PATH=$PATH:$PASH_TOP
+
+          $PASH_TOP/scripts/setup-annotations.sh
+          $PASH_TOP/scripts/ci/test_local_annotations.sh
```
angelhof, line 51: You probably need to save the exit code here and return it at the end, otherwise how does the CI know if the test failed?

compiler/orchestrator_runtime/pash_init_setup.sh
```diff
@@ -23,10 +23,10 @@ export pash_daemon_communicates_through_unix_pipes_flag=0
 export pash_speculative_flag=0
 export show_version=0
 export distributed_exec=0
+export local_annotations=0
```
angelhof, line 26: Could you use the naming convention from above and move the flag close to the other ones? `pash_checking_local_annotations`

docs/local-annotations-library-documentation.md
```diff
@@ -0,0 +1,69 @@
+# PaSh Setup and Execution Guide
+
+This guide walks you through **setting up and running PaSh**, including:
```
angelhof, line 3: We can make this a bit more specific to clarify that this walks you through setting up pash with local annotations

scripts/ci/test_local_annotations.sh
```diff
@@ -0,0 +1,74 @@
...
+    sed -i 's/^\(\s*\)# "cat": "Cat",/\1"cat": "Cat",/' "$FILE"
+    echo "Annotation restored after failure."
+    
+    exit 1
+fi
+
+# --------------------------------
+# Restore original cat annotation
+# --------------------------------
+sed -i 's/^\(\s*\)# "cat": "Cat",/\1"cat": "Cat",/' "$FILE"
```
angelhof, line 63: Maybe it is better to just save a copy of the file before the original sed and copying that here? It is easier to miss a bug like this :)

angelhof (commented):
Thanks a lot :) This is great! There is only one final thing missing (checking the exit code in CI) and some mini edits for clarity!

I also checked one of the latest PR runs in your repo and it seems that there are some issues (https://github.com/DivikChotani/pash/actions/runs/14053084765/job/39346886888#step:4:19). Are these warnings expected?

DivikChotani:
@angelhof Hi professor, I've made the changes requested to the documentation. I also saw you had some concerns about some of the warnings I get when running the script, just wanted to assure you those are expected, I just double checked and for the first run of pash on my device I get that warning as well, I also think that is the reason the POSIX correctness test fails. Thank you so much!

compiler/orchestrator_runtime/pash_init_setup.sh
```diff
@@ -23,7 +23,7 @@ export pash_daemon_communicates_through_unix_pipes_flag=0
 export pash_speculative_flag=0
 export show_version=0
 export distributed_exec=0
-export local_annotations=0
+export pash_checking_local_annotations=0
```
angelhof, line 26: Final nits: Could you move it with the other pash_checking flags?

scripts/ci/test_local_annotations.sh
```diff
@@ -60,7 +64,8 @@ fi
 # --------------------------------
 # Restore original cat annotation
 # --------------------------------
-sed -i 's/^\(\s*\)# "cat": "Cat",/\1"cat": "Cat",/' "$FILE"
+cp "AnnotationGenerationCopy.py" "$FILE"
+rm AnnotationGenerationCopy.py
```
angelhof, line 68: Final nits: You could also do `mv` here.

angelhof (approved):
Great, could you do these two final nits and we should be good to merge! Ping me here once you do them and I will merge. FYI the POSIX correctness issue doesn't have anything to do with your changes. These warnings also seem like they have to do with the python version.

DivikChotani:
@angelhof All done with the final nits, thanks!

PR #786, opened 2026-02-09
feat:  direct S3->Lambda streaming with boundary-aware chunking (1.2x to 7x speedup and high cost reduction)

victor-canard-duchene:
TLDR: Direct streaming from S3 to lambdas in parallel, bypassing EC2 splitter nodes, with boundary-aware chunking strategies; and smarter benchmarking/analysis scripts.

What was done:
  - Moved input path from EC2 splitter nodes to direct S3 range reads in Lambda workers.
  - Implemented and compared multiple S3 reader strategies (basic, dynamic/adaptive windows, approximate-boundaries)
  - Lambdas use headers to mimick r_split functionality but do not use headers when it is called with the -r flag
    - Adjustable`chunks_per_lambda` param to exploit more lambda parallelism and overall script pipelining
    - Preserve correctness by comparing outputs to leash (works for sort.sh through defining the same locale everywhere)  

  Performance progression (Unix50/6.sh, 20GB)
  - Leash baseline: 4m00s
  - Early “S3 optimized” basic attempts: 6m-7m (regression)
  - With eager dgsh-tee: 5m30s
  - Smart boundaries + direct streaming: ~3m00s (w=16)
      -> +Wider parallelism: 2m30s (w=32)
      -> +Tuned window sizing: 2m10s -> ~2m00s
  - Approximate boundaries + runtime correction: 1m10s

  Net impact
  - 4m00s -> 1m10s
  - End-to-end speedup: ~3.4x

  Key findings
  - FIFO `cat` sequencing can serialize work and hide parallelism.
  - Aggressive eager buffering can cause memory pressure/swapping at high width, but eager reading is crucial for speedup/ avoding timeouts on large inputs (I added eager nodes after the s3 reads on the lambdas; and left the eager nodes before each `r_merge` and `sort -m` nodes on the merger nodes)

Files:
Most of the work was done in `ir_helper.py` to adjust the IR (AST) before it is transformed to scripts that will run on EC2 and the lambdas

- `evaluation/benchmarks/run-leash-benchmark-matrix.sh`: benchmark matrix runner; supports `--noopt` and other flags for choosing which mode to run,
  - `evaluation/benchmarks/unix50/run-leash-benchmark-matrix.sh`: unix50 wrapper that forwards to the canonical runner above.
  - `evaluation/benchmarks/plot_results.py`: generates `grid_plot.png` from a benchmark `results.csv`.
  - `compiler/definitions/ir/nodes/serverless_remote_pipe.py`: strategy->reader-script
  mapping. this code makes the lambda calling code and its args depending on the mode selected (this is propagated through env vars set by the benchmark script)

  - `aws/s3-chunk-reader-smart-prealigned.py`: pre-aligned boundary reader (EC2 computes exact boundaries).
  - `aws/s3-chunk-reader-approx-tail-coord.py`: approximate boundaries + tail coordination (legacy/experimental) (similar to the DFR in DISH, read about it in docs/distributed-file-reader-reference.md.
  - `aws/s3-chunk-reader-approx-correction.py`: approximate boundaries + in-lambda correction (preferred).
  
  - `docs/S3_DIRECT_STREAMING_METHODS.md`: single source of truth for S3 methods, strategy mapping, and tuning.
  - `docs/LEASH_ANALYSIS.md`: comparative analysis across methods and tradeoffs.
  - `docs/S3_READER_REFACTOR_CHANGELOG.md`: history of S3 reader refactor and naming.
  
  How To Run (benchmark matrix)
per benchmark config is `leash-matrix-inputs.sh` where you define the inputs to the benchmark you want to run

  ```bash
  cd evaluation/benchmarks

  # Example: unix50 large input, selected modes
  ./run-leash-benchmark-matrix.sh --benchmark unix50 --large \
    --noopt --smart-prealigned --approx-dynamic --approx-adaptive-single-shot \
    --repeats 1

  # Or via benchmark-local wrapper
  cd unix50
  ./run-leash-benchmark-matrix.sh --large --noopt --approx-dynamic

  How To Plot

  python3 evaluation/benchmarks/plot_results.py \
    evaluation/benchmarks/unix50/benchmark_results/<timestamp>/results.csv

  (Produces grid_plot.png next to the CSV.)


### Testing 
The run-leash-benchmark-matrix.sh script downloads and compares the outputs of each mode with normal leash to make sure this optimization isn't breaking anything. It also retrieves the time and cost so we can compare those with leash and pash. 
Finally, I implemented this optimization after doing a lot of testing to see the transfer rates from s3 to ec2 and from ec2 to lambdas (this is documented and available in testing/s3_transfer_benchmarks, along with a README to explain how to re-run these initial tests).
Note: A lot of other testing was done but is only present on my local machine so as to not clutter the repo


compiler/definitions/ir/nodes/dgsh_tee.py
```diff
@@ -16,20 +16,38 @@ def __init__(self,
...
+    #6.sh LEASH OOM with -I flag 
+    #TODO: tune it -m 
+    
+
+    if is_s3:
+        flag_option_list = [OptionWithIO("-i", input_id),
                         OptionWithIO("-o", output_id),
                         Flag("-I"),
-                        # Flag("-f"),
+                        OptionWithIO("-m", ArgStringType(Arg.string_to_arg("2G"))),
```
yizhengx, line 35: Is this branch making dgsh-tee nodes for lambdas running s3-get? If so, would `-m 2G` cause OOM given the default lambda size is 1769M?

aws/legacy/lambda.zip
yizhengx, line 1: Would it make sense to remove zip and binary?

compiler/serverless/ir_helper.py
```diff
@@ -175,15 +496,210 @@ def add_nodes_to_subgraphs(subgraphs:List[IR], file_id_gen: FileIdGen, input_fif
...
+                                                                                job_uid=job_uid,
+                                                                                skip_first_line=skip_first_line,
+                                                                                window_size=window_size_param,
+                                                                                chunks_per_lambda=chunks_per_lambda_param,
+                                                                                window_after_vec=window_after_vec_param,
+                                                                                write_headers=write_headers,
+                                                                                s3_reader_strategy=s3_reader_strategy)
+                        
+                        # s3getobj byterange=start-end -> sort   : lambda1
+                        lambda_counter += 1
```
yizhengx: This relies on the assumption that the iteration of subgraphs follows the order in merging them right? This is likely true in all cases but I'm not sure if it's possible this can be false.

compiler/serverless/ir_helper.py
```diff
@@ -175,15 +496,210 @@ def add_nodes_to_subgraphs(subgraphs:List[IR], file_id_gen: FileIdGen, input_fif
...
+                    else:
+                        remote_read = serverless_remote_pipe.make_serverless_remote_pipe(local_fifo_id=ephemeral_edge.get_ident(),
+                                                                                is_remote_read=True,
+                                                                                remote_key=filename,
+                                                                                output_edge=None,
+                                                                                is_tcp=False)
+                    if in_edge == ec2_in_edge and not args.no_eager:
```
yizhengx, line 699: I'm not sure if this is necessary because we should have dgsh per tcp receiver (on ec2) to eagerly pull output of lambda, which should avoid blocking lambda processing from slower consumer. The concern I have is if the lambda processing itself is slow, this eager will create memory/disk pressure for lambdas.

docs/S3_DIRECT_STREAMING_METHODS.md
```diff
@@ -0,0 +1,196 @@
...
+## How Strategy Selection Works
+
+There are two paths:
+
+1. Preferred path: `ir_helper.py` computes and passes `s3_reader_strategy` explicitly.
+2. Backward-compatible path: `serverless_remote_pipe.py` derives strategy from environment flags.
+
+Environment fallback behavior:
+
+- Any of these flags set to `true` selects `s3_approx_correction`:
```
yizhengx, line 40: Could you also document the difference between these correction modes? (Maybe they are documented somewhere but I couldn't find them)

yizhengx:
Thanks Victor! These changes look good, I just left a few minor comments.

compiler/definitions/ir/nodes/dgsh_tee.py
victor-canard-duchene: Yes that is true but I bumped my lambda's up to 3GB. We can adjust this to a smaller limit for smaller memory lambdas. 
For example we can make this 1G for the default lambdas, the idea is just so it pulls as much as possible. We will need to re-add the -f flag though so it go OOM and can store the excess on disk.

aws/legacy/lambda.zip
victor-canard-duchene: Yes I removed it

compiler/serverless/ir_helper.py
victor-canard-duchene: That concern is valid. 
With the way that `r_merge` currently orders on input position this only works because the order of subgraph iteration in this loop is the same as the rmerge input order.
I tested it with a shuffled order and it didn't work.

Possible solutions:
1. Easiest. We assume that our code will stay the same, thus, we always iterate over lambda subgraphs in the same order as we place them as inputs of r_merge
2. More involved but more correct. Save the order of the rmerge inputs and assign lambda counter values accordingly to which rmerge input they correspond to. This was tested and it works.

yizhengx: I think it's okay to move forward with the easiest solution now but let's keep in mind that this could possibly go wrong.

compiler/definitions/ir/nodes/dgsh_tee.py
yizhengx: Could you do a simple testing to see if adding dgsh-tee on lambda make a big diff? E.g., run the same exp with and without this. You might have done sth similar but just to see if this is actually needed.

compiler/serverless/ir_helper.py
victor-canard-duchene: I think that even when lambda-side processing is slow, we should still read from S3 as aggressively as possible. Otherwise, once one chunk finishes, the lambda may idle while waiting for the next S3 fetch, especially since FIFO capacity is limited (around 64 KiB).

To avoid memory pressure, the eager buffer cap (-m) should be sized relative to lambda memory. For example, -m 2G may be too high for a 1.7 GB lambda, where something like -m 500M would be more appropriate.

In other configurations (for example, smart prealigned boundaries computed on EC2 with CHUNKS_PER_LAMBDA=1), these eager nodes reduced runtime from about 7 minutes to about 5 minutes 30 seconds.

  So I think we should keep these eager nodes: they appear safe when tuned correctly, and they provide substantial speedups in low-chunk scenarios and smaller but still positive gains in high-chunk scenarios.

  You’re right that we already have eager pulls before merger stages (sort -m, r_merge, etc.), but that is a different part of the pipeline. These input-side eager nodes shift the bottleneck from S3 -> consumer to writer  -> consumer. That tends to help s3_smart_prealigned more than s3_approx_correction, but it can still help both.

compiler/serverless/ir_helper.py
victor-canard-duchene: 2. was implemented and I pushed it, just in case we implement other things/features that break this implicit order.

victor-canard-duchene: Because indeed they do rely on this order. I tested shuffling the graphs which broke everything until I added this explicit r_split output id map to lambda id.

docs/S3_DIRECT_STREAMING_METHODS.md
victor-canard-duchene: I've added a section about this in the same doc and pushed it. Let me know if it's detailed enough.

Issue #733, opened 2024-11-18
Understanding and reducing compilation overheads


angelhof:
The compiler seems to have non negligible overhead when dealing with large graphs (for very large widths, e.g., 200). It would be good to understand these overheads and determine if they can be reduced. 

I would try to compile the nfa-regex script (https://github.com/binpash/pash/blob/main/evaluation/benchmarks/oneliners/nfa-regex.sh) with increasing widths to estimate overheads. The `--dry_run_compiler` flag could be used to not execute the compiled script but still compile it (it is pretty old so I am not sure if it still works).

The first step would be to understand which method calls of the compiler contribute mostly to the compilation overhead. The next step would be to reimplement some of them to minimize such overheads.

PR #787, opened 2026-04-17
Issue#733 reduce compilation overhead

overlorde:
Hello All, I am trying to solve this issue #733 , please let me know about the modifications i need to do / feedbacks. 
I tried to sovle ir.py:546 comments through directly constructing shasta types and bypassing deepcopy to gain some performance boost. 

 - Added cProfile around `compile_ir()` in `compilation_server.py` to bottlenecks - round trips `ir.py:547`, `ir_to_ast.py:114`, `ir_to_ast.py:151` 
 - Replace `make_kv("Tag", [...])` dict construction + `to_ast_node(...)` round-trips with direct shasta typed constructors                                                                                                                   
  - Bypass `CommandInvocationWithIOVars.__init__` deepcopy at 8 call sites via `__new__` around ~100 ms reduction on top of shasta type                                                                         
  - **Result:** 4-6x faster compilation at width=200 on heavy one-liners (e.g., `nfa-regex` drops from 2675ms to 421ms), and                          
  - `evaluation/tests/test_evaluation_scripts.sh` passes 100/100.
  - Bypass `__init__` is safe — no caller mutates these containers after passing them in. Proper fix is to remove the `deepcopy` in `pash-annotations` itself.
  
speedup gained for nfa-regex.sh
### nfa-regex.sh — compile time (`--dry_run_compiler`, 3-run avg)
  
   | Width | Original (ms) | PR (ms) | Speedup |
   |-------|---------------|---------|---------|
   | 2     | 42.8          | 10.2    | 4.21x   |
   | 16    | 218.9         | 36.3    | 6.02x   |
   | 64    | 834.4         | 132.7   | 6.29x   |
   | 200   | 2674.9        | 421.4   | 6.35x   |

### DIFF
--- a/src/pash/compiler/ir.py
+++ b/src/pash/compiler/ir.py
@@ -541,8 +541,6 @@ class IR:
             assignment = self.collect_pid_assignment()
             asts.append(assignment)
 
-        ## TODO: Ideally we would like to make them as typed nodes already
-        ## Solved: returns are shashta typed
         return asts
 
     def collect_pid_assignment(self):
--- a/src/pash/compiler/ir.py
+++ b/src/pash/compiler/ir.py
@@ -22,8 +22,6 @@ from pash_annotations.datatypes.CommandInvocationWithIOVars import (
     CommandInvocationWithIOVars,
 )
 
-from shasta.ast_node import ast_node_to_untyped_deep
-from shasta.json_to_ast import to_ast_node
 
 from annotations_utils.util_parsing import parse_arg_list_to_command_invocation
 from annotations_utils.util_cmd_invocations import (
@@ -544,8 +542,8 @@ class IR:
             asts.append(assignment)
 
         ## TODO: Ideally we would like to make them as typed nodes already
-        class_asts = [to_ast_node(ast_node_to_untyped_deep(ast)) for ast in asts]
-        return class_asts
+        ## Solved: returns are shashta typed
+        return asts
 
     def collect_pid_assignment(self):
         ## Creates:
--- a/src/pash/compiler/ir.py
+++ b/src/pash/compiler/ir.py
@@ -247,6 +247,7 @@ def add_file_id_vars(command_invocation_with_io, fileIdGen):
     )
     """
 
+    # Skip __init__ to avoid its deepcopy; inputs are freshly constructed here.
     cmd_inv_with_io_vars = CommandInvocationWithIOVars.__new__(CommandInvocationWithIOVars)
     cmd_inv_with_io_vars.cmd_name = command_invocation_with_io.cmd_name
     cmd_inv_with_io_vars.flag_option_list=new_flagoption_list
--- a/src/pash/compiler/ir.py
+++ b/src/pash/compiler/ir.py
@@ -236,7 +236,7 @@ def add_file_id_vars(command_invocation_with_io, fileIdGen):
         )
     else:
         new_implicit_use_of_streaming_output = None
-
+    """
     command_invocation_with_io_vars = CommandInvocationWithIOVars(
         cmd_name=command_invocation_with_io.cmd_name,
         flag_option_list=new_flagoption_list,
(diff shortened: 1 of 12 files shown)


src/pash/compiler/ir.py
```diff
@@ -544,8 +555,8 @@ def to_ast(self, drain_streams) -> "list[AstNode]":
             asts.append(assignment)
 
         ## TODO: Ideally we would like to make them as typed nodes already
```
angelhof, line 557: Remove this comment and the solved now that it is OK :)

src/pash/compiler/ir_defs/nodes/pash_split.py
```diff
@@ -37,6 +37,7 @@ def make_split_file(input_id, out_ids):
     operand_list.extend(out_ids)
     access_map = {output_id: make_stream_output() for output_id in out_ids}
     access_map[input_id] = make_stream_input()
+    """
```
angelhof, line 40: Completely remove this code instead of commenting it out! (similarly in all other places!

src/pash/compiler/ir_defs/nodes/pash_split.py
```diff
@@ -45,4 +46,16 @@ def make_split_file(input_id, out_ids):
         implicit_use_of_streaming_output=None,
         access_map=access_map,
     )
+    """
+
+
+    # Skip __init__ to avoid its deepcopy; inputs are freshly constructed here.
```
angelhof, line 52: Is there any other way to avoid the deepcopy without avoiding the standard init? This seems a bit janky to me to not use Python's normal object initialization. There must be a way to do this

angelhof (changes requested):
Great work on these changes! Can you fix/respond to my points? In particular I would really like to avoid not using the normal Python object initialization. There must be a way to avoid the deepcopy there right?

src/pash/compiler/ir_defs/nodes/pash_split.py
overlorde: @angelhof one way on top of my mind is runtime modification of the `__init__` , replacing it with `fast_init` in `pash`. The existing initialization codes around the pash repo would stay the same.  To have a more stable fix, need to go into the pash annotations repo, but the scope becomes larger then, and side effects needs to be checked. Runtime modification / monkeypatching seems like a way where we can use pythons normal object initialization with minimal code changes. I am adding  modifications in a while.

angelhof: I think the proper fix requires modifying the pash annotations repo, which keeps the scope lower than bypassing Python's init mechanism :) I think we should only accept a fix that requires bypassing init if the gains are on the critical path and very very significant (compilation is not on the critical path, so improving it but making maintenance harder is not worth it IMO).

overlorde: ### nfa-regex.sh — compile time (`--dry_run_compiler`, 3-run avg - with _fast_init)                                                                                  
                                                                                                                                                     
  | Width | Original (ms) | PR (ms) | Speedup |                                                                                                      
  |-------|---------------|---------|---------|
  | 2     | 42.33         | 6.12    | 6.92x   |                                                                                                      
  | 16    | 214.51        | 5.92    | 36.23x  |                                                                                                      
  | 64    | 827.98        | 6.95    | 119.13x |
  | 200   | 2615.91       | 7.32    | 357.37x |
  
  The speedup is way more than before because of the removal of deepcopy from `_init_` - bypassing with `_fast_init_` - this time its global cause all instances of `__init__` was modified during start/runtime. Same modification on pash annotations repo should also give the same result. As per the last comment, proper fix would be modifying the pash annotations repo, which I agree.

overlorde: reverted back to the original init - with shasta typed objects, comments are addressed. Need an integrated test to check both.

overlorde:
Back to original `__init__`, comments addressed.

angelhof:
Great, do we get any speedups now? If not, what is the expected path to get back the deepcopy speedup? What needs to happen in the annotation repo?

overlorde:
> Great, do we get any speedups now? If not, what is the expected path to get back the deepcopy speedup? What needs to happen in the annotation repo?

Yes, we still get speedups from converting to shasta typed objects, below are the ones: 

Script                   |   Width | Orig    |  PR       | Speedup
nfa-regex.sh       |        2    |   42.7   |    8.5    |  5.03x
nfa-regex.sh       |      16    | 215.3   |   43.8   |   4.92x
nfa-regex.sh       |      64    | 828.4   | 150.1   |     5.52x
nfa-regex.sh       |     200   | 2626.7 |    510.9|    5.14x

The deepcopy speedup is more. 
The expected path to get deepcopy speedup would be modifying `CommandInvocationWithIOVars.py` from `annotations` repo : line 26 to 30 - turn to shallow copy from deepcopy - except for `streaming input/outputs` - they can be directly assigned. That would lead to rethink how regression tests would work out for the annotations repo/need to write test cases for new modifications.

PR #790, opened 2026-09-22
Strip syntactic quotes from command names before annotation lookup

### LINKED ISSUE #741: Quoted commands aren’t recognized as being parallelizable.

ermi9:
Annotations are keyed by executable name (cat), but format_arg_chars regenerates shell syntax and re-emits quote characters, so a quoted command name was looked up as "cat" and never matched. The region was then treated as side-effectful and run sequentially.

Add remove_quotes_expanded_arg_chars, which walks the arg_char list and unwraps QArgChar nodes rather than manipulating the formatted string, so nested and partial quoting is handled correctly. Words containing anything other than C, E or Q chars are not fully expanded, so their value is unknown; these return None and the caller falls back to the previous behaviour.

Fixes #741


src/pash/compiler/util.py
```diff
@@ -120,6 +121,22 @@ def format_arg_chars(arg_chars):
     return "".join(chars)
 
 
+def remove_quotes_expanded_arg_chars(arg_chars) -> Optional[str]:
```
angelhof: Can you add a little comment describing what this function does/its assumptions here?

src/pash/compiler/annotations_utils/util_parsing.py
```diff
@@ -65,7 +65,10 @@ def fix_parsing_newline(arg):
 def parse_arg_list_to_command_invocation(
     command, flags_options_operands
 ) -> CommandInvocationInitial:
-    cmd_name = format_arg_chars(command)
+    cmd_name = remove_quotes_expanded_arg_chars(command)
+    if cmd_name is None:
```
angelhof: Similarly, adding a comment here that this is an optimistic check that only allows Q, E, C and fails for other chars

angelhof (commented):
Looks good modulo the changes I proposed. Unfortunately, the tests seem to fail now (due to node 20 being obsolete). We can't merge before passing tests, so either we will wait until I have time to check this myself, or you (or someone else) can attempt to solve the CI node issue in another PR so that we can get things unstuck :)

